PR #81, opened 2020-03-15
Add current PHP versions to Travis CI configuration

fredden:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description
PHP version constraint in `composer.json` permit usage with PHP versions 7.3 & 7.4, but these versions do not run the testsuite via Travis CI. Additionally, it seems sensible to run codesniffer & coveralls on the latest support version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,18 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
+    - php: 7.3
+      env:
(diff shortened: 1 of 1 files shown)


michalbundyra:
Thanks, @fredden!

PR #82, opened 2020-03-21
Storage\Maildir fixes + enable unit tests

fredden:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description
Increase test coverage. While test coverage is low, the GitHub status check from coverage is reporting failure on every pull request. While this change alone is unlikely to bring coverage above the threshold, it'll help.

### DIFF
--- a/src/Storage/Folder/Maildir.php
+++ b/src/Storage/Folder/Maildir.php
@@ -97,7 +97,7 @@ class Maildir extends Storage\Maildir implements FolderInterface
                 continue;
             }
 
-            if ($this->_isMaildir($this->rootdir . $entry)) {
+            if ($this->isMaildir($this->rootdir . $entry)) {
                 $dirs[] = $entry;
             }
         }
@@ -187,7 +187,7 @@ class Maildir extends Storage\Maildir implements FolderInterface
         $folder = $this->getFolders($this->currentFolder);
 
         try {
-            $this->_openMaildir($this->rootdir . '.' . $folder->getGlobalName());
+            $this->openMaildir($this->rootdir . '.' . $folder->getGlobalName());
         } catch (Exception\ExceptionInterface $e) {
             // check what went wrong
             if (! $folder->isSelectable()) {
--- a/src/Storage/Folder/Maildir.php
+++ b/src/Storage/Folder/Maildir.php
@@ -97,7 +97,7 @@ class Maildir extends Storage\Maildir implements FolderInterface
                 continue;
             }
 
-            if ($this->isMaildir($this->rootdir . $entry)) {
+            if ($this->_isMaildir($this->rootdir . $entry)) {
                 $dirs[] = $entry;
             }
         }
@@ -187,7 +187,7 @@ class Maildir extends Storage\Maildir implements FolderInterface
         $folder = $this->getFolders($this->currentFolder);
 
         try {
-            $this->openMaildir($this->rootdir . '.' . $folder->getGlobalName());
+            $this->_openMaildir($this->rootdir . '.' . $folder->getGlobalName());
         } catch (Exception\ExceptionInterface $e) {
             // check what went wrong
             if (! $folder->isSelectable()) {
--- a/src/Storage/Maildir.php
+++ b/src/Storage/Maildir.php
@@ -133,7 +133,7 @@ class Maildir extends AbstractStorage
     public function getMessage($id)
     {
         // TODO that's ugly, would be better to let the message class decide
-        if ($this->messageClass === Message\File::class
+        if (\trim($this->messageClass, '\\') === Message\File::class
             || is_subclass_of($this->messageClass, Message\File::class)
         ) {
             return new $this->messageClass([
@@ -336,6 +336,9 @@ class Maildir extends AbstractStorage
             }
             $this->files[] = $data;
         }
+        \usort($this->files, function ($a, $b) {
+            return \strcmp($a['filename'], $b['filename']);
+        });
     }
 
     /**
--- a/src/Storage/Maildir.php
+++ b/src/Storage/Maildir.php
@@ -133,7 +133,7 @@ class Maildir extends AbstractStorage
     public function getMessage($id)
     {
         // TODO that's ugly, would be better to let the message class decide
-        if (\trim($this->messageClass, '\\') === Message\File::class
+        if ($this->messageClass === Message\File::class
             || is_subclass_of($this->messageClass, Message\File::class)
         ) {
             return new $this->messageClass([
--- a/src/Storage/Maildir.php
+++ b/src/Storage/Maildir.php
@@ -133,7 +133,7 @@ class Maildir extends AbstractStorage
     public function getMessage($id)
     {
         // TODO that's ugly, would be better to let the message class decide
-        if ($this->messageClass === Message\File::class
+        if (\trim($this->messageClass, '\\') === Message\File::class
             || is_subclass_of($this->messageClass, Message\File::class)
         ) {
             return new $this->messageClass([
--- a/test/AddressListTest.php
+++ b/test/AddressListTest.php
@@ -165,4 +165,39 @@ class AddressListTest extends TestCase
         $this->assertTrue($addressList->has('asda.fasd@example.net'));
         $this->assertTrue($addressList->has('root@example.org'));
     }
+
+    public function testMergeTwoLists()
+    {
+        $otherList = new AddressList();
+        $this->list->add('one@example.net');
+        $otherList->add('two@example.org');
+        $this->list->merge($otherList);
+        $this->assertEquals(2, count($this->list));
+    }
+
+    public function testDeleteSuccess()
+    {
+        $this->list->add('test@example.com');
+        $this->assertTrue($this->list->delete('test@example.com'));
+        $this->assertEquals(0, count($this->list));
+    }
+
+    public function testDeleteNotExist()
+    {
+        $this->assertFalse($this->list->delete('test@example.com'));
+    }
+
+    public function testKey()
+    {
+        $this->assertNull($this->list->key());
+        $this->list->add('test@example.com');
+        $this->list->add('test@example.net');
+        $this->list->add('test@example.org');
+        $this->list->rewind();
+        $this->assertSame('test@example.com', $this->list->key());
+        $this->list->next();
+        $this->assertSame('test@example.net', $this->list->key());
+        $this->list->next();
+        $this->assertSame('test@example.org', $this->list->key());
+    }
 }
--- a/test/AddressListTest.php
+++ b/test/AddressListTest.php
@@ -40,6 +40,12 @@ class AddressListTest extends TestCase
 
     public function testAddingEmailsIncreasesCount()
     {
+        $this->list->add('test@example.com');
+        $this->assertEquals(1, count($this->list));
+    }
+
+    public function testAddingEmailFromStringIncreasesCount()
+    {
         $this->list->addFromString('test@example.com');
         $this->assertEquals(1, count($this->list));
     }
--- a/test/AddressListTest.php
+++ b/test/AddressListTest.php
@@ -40,7 +40,7 @@ class AddressListTest extends TestCase
 
     public function testAddingEmailsIncreasesCount()
     {
-        $this->list->add('test@example.com');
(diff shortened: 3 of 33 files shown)


test/Header/IdentificationFieldTest.php
```diff
@@ -68,4 +77,44 @@ public function testSerializationToString($className, $headerString, $ids)
...
         $this->assertEquals($headerString, $header->toString());
     }
+
+    /**
+     * @dataProvider stringHeadersProvider
+     * @param string $className
+     * @param string $headerString
+     * @param string[] $ids
+     */
+    public function testEncodingAccessors($className, $headerString, $ids)
```
michalbundyra: please split into two separate tests as suggested before

also here: we can add typehint `array` for `$ids`

I see this type of test several more below, all should be splitted into two separate tests, imo.

test/Header/IdentificationFieldTest.php
```diff
@@ -68,4 +77,44 @@ public function testSerializationToString($className, $headerString, $ids)
...
+    }
+
+    /**
+     * @dataProvider invalidIds
+     * @param string $className
+     * @param string[] $ids
+     */
+    public function testSetIdsThrowsOnInvalidInput($className, $ids)
+    {
+        $this->expectException('Laminas\Mail\Header\Exception\InvalidArgumentException');
```
michalbundyra: please move expectException just before the call which should result in the exception.

From this test we do not know if the exception is thrown in the constructor or in `setIds` call.

test/Header/MessageIdTest.php
```diff
@@ -35,6 +37,15 @@ public function testAutoGeneration()
         $this->assertContains('@', $messageid->getFieldValue());
     }
 
+    public function testAutoGenerationWithServerVars()
+    {
+        $_SERVER['REMOTE_ADDR'] = '172.16.0.1';
+        $_SERVER['SERVER_NAME'] = 'server-name.test';
```
michalbundyra: We need to make sure that changing global variables is not propagated between tests.

test/Header/MessageIdTest.php
```diff
@@ -35,6 +37,15 @@ public function testAutoGeneration()
...
+        $messageid = new Header\MessageId();
+        $messageid->setId();
+
+        $this->assertContains('@', $messageid->getFieldValue());
```
michalbundyra: I am not sure what we are testing here, can you explain, please?

test/MessageTest.php
```diff
@@ -471,6 +471,7 @@ public function testSubjectIsEmptyByDefault()
 
     public function testSubjectIsMutable()
     {
+        $this->message->setSubject('test subject');
```
michalbundyra: Can you explain this addition, please?

test/Storage/MaildirFolderTest.php
```diff
@@ -365,7 +365,7 @@ public function testGetVanishedFolder()
         $root = $mail->getFolders();
         $root->foobar = new Folder('foobar', 'foobar');
 
-        $this->expectException('Laminas\Mail\Storage\Exception\InvalidArgumentException');
+        $this->expectException('Laminas\Mail\Storage\Exception\RuntimeException');
```
michalbundyra: Can you explain this change and all below, please?

test/Storage/MaildirMessageOldTest.php
```diff
@@ -94,23 +94,23 @@ public function testFetchHeader()
 /*
     public function testFetchTopBody()
     {
-        $mail = new MaildirOldMessage(array('dirname' => $this->maildir));
+        $mail = new TestAsset\MaildirOldMessage(array('dirname' => $this->maildir));
```
michalbundyra: As the file name suggest "Old" maybe this file should be deleted? Can you see any value of these tests?

test/Storage/MaildirTest.php
```diff
@@ -216,7 +216,7 @@ public function testFailedRemove()
     {
         $mail = new Storage\Maildir(['dirname' => $this->maildir]);
 
-        $this->expectException('Laminas\Mail\Storage\Exception\InvalidArgumentException');
+        $this->expectException('Laminas\Mail\Storage\Exception\RuntimeException');
```
michalbundyra: Can you explain all these changes, please?

src/Storage/Folder/Maildir.php
```diff
@@ -97,7 +97,7 @@ protected function buildFolderTree()
                 continue;
             }
 
-            if ($this->_isMaildir($this->rootdir . $entry)) {
+            if ($this->isMaildir($this->rootdir . $entry)) {
```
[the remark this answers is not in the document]
froschdesign: > Unless there is a magic call somewhere?

No magic call here.

@fredden 
This is an bugfix and should be send as separate pull request:

https://github.com/laminas/laminas-mail/blob/ab5d76a92a10fa33a198c08149bba48faf37b0bd/src/Storage/Maildir.php#L240-L246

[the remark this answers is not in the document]
fredden: I'll move this to a separate pull request. I'm working on the Maildir tests as a large portion of them currently fail when enabled.

src/Storage/Folder/Maildir.php
```diff
@@ -187,7 +187,7 @@ public function selectFolder($globalName)
         $folder = $this->getFolders($this->currentFolder);
 
         try {
-            $this->_openMaildir($this->rootdir . '.' . $folder->getGlobalName());
+            $this->openMaildir($this->rootdir . '.' . $folder->getGlobalName());
```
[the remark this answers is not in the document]
fredden: Do we have tests: yes. They're disabled by default at present. I'm working on these separately. I'll move this bugfix to that branch.

src/Storage/Maildir.php
```diff
@@ -133,7 +133,7 @@ public function getSize($id = null)
     public function getMessage($id)
     {
         // TODO that's ugly, would be better to let the message class decide
-        if ($this->messageClass === Message\File::class
+        if (\trim($this->messageClass, '\\') === Message\File::class
```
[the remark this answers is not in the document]
fredden: Do we have tests: yes. They're disabled by default at present. I'm working on these separately. I'll move this bugfix to that branch.

src/Storage/Folder/Maildir.php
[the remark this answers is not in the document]
froschdesign: @fredden 
Many thanks in advance! 👍

test/Header/ContentTypeTest.php
```diff
@@ -187,4 +187,54 @@ public function invalidHeaderLinesProvider()
...
+    public function testSetEncoding()
+    {
+        $header = new ContentType('today');
+        $header->setEncoding('UTF-8');
+        $this->assertSame('UTF-8', $header->getEncoding());
+    }
+
+    public function testSetTypeThrowsOnInvalidValue()
+    {
+        $this->expectException('Laminas\Mail\Header\Exception\InvalidArgumentException');
```
michalbundyra: Please move before `setType` call.

Like that it's not clear where the exception is thrown - in the constructor or in the setter.

test/Header/GenericHeaderTest.php
```diff
@@ -16,22 +16,30 @@
  */
 class GenericHeaderTest extends TestCase
 {
+    public function invalidHeaderLines()
+    {
+        return [
+            'appned-chr-32' => ['Content-Type' . chr(32) . ': text/html'],
```
michalbundyra: typo in key: "append"

test/Header/ReceivedTest.php
```diff
@@ -90,4 +90,55 @@ public function testConstructorRaisesExceptionOnValueWithCRLFInjectionAttempt($v
...
+        $this->assertEquals(
+            "Received: test\r\nReceived: test2\r\nReceived: test3",
+            $header->toStringMultipleHeaders([$header2, $header3])
+        );
+    }
+
+    public function testToStringMultipleHeadersThrows()
+    {
+        $this->expectException('Laminas\Mail\Header\Exception\RuntimeException');
+        $this->expectExceptionMessage('can only accept an array of Received headers');
```
michalbundyra: these two call should be moved after class instantiation, so we know that the exception is thrown from `toStringMultipleHeaders` call.

test/AddressListTest.php
```diff
@@ -40,7 +40,7 @@ public function testIsEmptyByDefault()
 
     public function testAddingEmailsIncreasesCount()
     {
-        $this->list->add('test@example.com');
+        $this->list->addFromString('test@example.com');
```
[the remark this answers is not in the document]
fredden: `addFromString()` calls `add()` internally, but it makes sense to have these as separate test. I've split this out.

test/Header/ContentTransferEncodingTest.php
```diff
@@ -147,4 +148,18 @@ public function testSetTransferEncodingRaisesExceptionForInvalidValues()
...
         $header->setTransferEncoding("8bit\r\n 7bit");
     }
+
+    public function testFromStringRaisesExceptionOnInvalidHeader()
+    {
+        $this->expectException('Laminas\Mail\Header\Exception\InvalidArgumentException');
+        ContentTransferEncoding::fromString('Foo: bar');
+    }
+
+    public function testEncodingAccessors()
```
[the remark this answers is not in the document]
fredden: Thanks; I've split this pattern out where I've used it into separate tests.

test/Header/ContentTypeTest.php
```diff
@@ -187,4 +187,34 @@ public function invalidHeaderLinesProvider()
...
+        $header->setType('invalid');
+    }
+
+    public function testGetRemoveParameter()
+    {
+        $header = ContentType::fromString('content-type: text/plain; level=top');
+        $this->assertEquals('top', $header->getParameter('level'));
+        $this->assertEquals(true, $header->removeParameter('level'));
+        $this->assertEquals(false, $header->removeParameter('level'));
+        $this->assertEquals(null, $header->getParameter('level'));
```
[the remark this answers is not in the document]
fredden: Thanks for the tip about `assertSame` versus `assertEquals`. I've split this into several tests using the suggested assertions.

test/Header/GenericHeaderTest.php
```diff
@@ -16,22 +16,30 @@
  */
 class GenericHeaderTest extends TestCase
 {
+    public function invalidHeaderLines()
+    {
+        return [
+            ['Content-Type' . chr(32) . ': text/html; charset = "iso-8859-1"' . "\nThis is a test"],
+            ['Content-Type: text/html; charset = "iso-8859-1"' . "\nThis is a test"],
```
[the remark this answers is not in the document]
fredden: The first string here is the original test. I've labelled these all now.

fredden:
Thanks for all the feedback. I've a few points to work through and will then request another review.

test/MessageTest.php
fredden: This line covers line 342 of `src/Message.php`, which is not touched by the existing test-suite. I've moved this to a separate test now.

test/Header/MessageIdTest.php
fredden: This is a copy of the preceding test. I've adjusted this line to include the hostname supplied on line 43 now.

test/Header/MessageIdTest.php
fredden: Thanks, they were indeed persisting. I've added a couple of lines to solve this now.

src/Storage/Maildir.php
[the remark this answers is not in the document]
michalbundyra: @fredden before I start reviewing it again just a quick question: you were saying about extracting bugfixes into separate PR - is it still a plan? I can still see all changes here. Thanks!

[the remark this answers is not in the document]
fredden: I was going to move all the maildir test fixes to a separate pull request, but decided to leave here. If I move the bugfixes to a separate pull request then the testsuite will fail here until that lands. (The tests were being skipped by default, but run now.)

I'm happy to go with your recommendation. (Moving to a separate pull request isn't a lot of work.)

michalbundyra:
Thanks, @fredden!

PR #83, opened 2020-03-27
Transport\InMemory: no messages could have been sent, fix return type

Slamdunk:
It may happen that no messages have been sent, so the return type could be null as well.

### DIFF
--- a/src/Transport/InMemory.php
+++ b/src/Transport/InMemory.php
@@ -20,7 +20,7 @@ use Laminas\Mail\Message;
 class InMemory implements TransportInterface
 {
     /**
-     * @var Message
+     * @var null|Message
      */
     protected $lastMessage;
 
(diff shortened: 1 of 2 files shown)


michalbundyra (changes requested):
@Slamdunk Just checked, it would be nice to add test case as well. `null` behaviour is currently untested. Thanks!

michalbundyra (approved):

Slamdunk:
Mmm, I don't get why DCO fails, my latest commit is signed-off:
immagine
Do I have to sign-off also the first commit?

michalbundyra:
@Slamdunk 

> Do I have to sign-off also the first commit?

In the current situation yes, but I've heard that "app update is on the way" to check signoff just on head commit. So if you can signoff the first commit as well, it would be great. See https://github.com/laminas/laminas-mail/pull/83/checks?check_run_id=544640351 for more details.

michalbundyra:
Thanks, @Slamdunk!

PR #89, opened 2020-05-11
Change $params type in PHPDoc of Storage\Imap|Pop3

chrisdeeming:
Signed-off-by: chrisdeeming <chris@xenforo.com>

|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

`Storage\Imap` and `Storage\Pop3` both support passing in a configured `Protocol` object instead of an array so adjust the PHPDoc to reflect that support.


weierophinney:
Thank you, @chrisdeeming!

PR #90, opened 2020-06-20
Use link to ZF tracker

glensc:
I was going over diffs to see what changed zend-mail 2.10.0 to laminas 2.10.1 and noticed a bad replace.

This fixes bad replace ZF->Laminas by using a link instead


michalbundyra:
Thanks, @glensc!

PR #91, opened 2020-06-20
Fix description of HeaderInterface constants

glensc:
Fixes nonsense that came after ZF->Laminas transition (what is Laminas value?). Well it was not understandable description already before.

### DIFF
--- a/src/Header/HeaderInterface.php
+++ b/src/Header/HeaderInterface.php
@@ -11,14 +11,14 @@ namespace Laminas\Mail\Header;
 interface HeaderInterface
 {
     /**
-     * Format value in Mime-Encoding if not US-ASCII encoding is used
+     * Format value in Mime-Encoding (Quoted-Printable). Result is valid US-ASCII string
      *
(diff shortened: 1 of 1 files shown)


src/Header/HeaderInterface.php
```diff
@@ -11,14 +11,14 @@
...
     /**
-     * Format value in Mime-Encoding if not US-ASCII encoding is used
+     * Format value in Mime-Encoding (Quoted-Printable). Result is valid US-ASCII string
      *
      * @var bool
      */
     const FORMAT_ENCODED = true;
 
     /**
-     * Return value with the interval Laminas value (UTF-8 non-encoded)
```
weierophinney, line 21: This "nonsense" was inherited - originally it read "interval ZF2 value". So please, don't blame it on the transition; as they say garbage-in-garbage-out.

glensc: yeah, I git blame'd and it leads to the initial commit.

PR #92, opened 2020-06-20
Use ::class constants instead of strings to class names

glensc:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

As this project minimum PHP version is [5.6], can use [::class] constant safely.

This improves consistency in the project (some places already use `::class`).

[5.6]: https://github.com/laminas/laminas-mail/blob/2.10.1/composer.json#L32
[::class]: https://www.php.net/manual/en/language.oop5.constants.php


glensc:
this is the only place I kept using string because likely syntax error otherwise:
```php
            $types[] = ['Laminas\Mail\Transport\Null'];
```

weierophinney:
Thanks, @glensc; merged to develop for release with 2.11.0.

Issue #14, opened 2019-12-31
Fix AddressList toString method to quote semicolon


weierophinney:
Certain input of `AddressList` headers, cannot be converted to string and back to header object because of incorrect quoting:

```
From: "Foo;" <root@example.com>
```

gets incorrectly converted as
```
From: Foo; <root@example.com>
```

but `;` is address separator, so it needs to be quoted:

```
From: "Foo;" <root@example.com>
```

The problem I discovered internally when using `Storage\Message` with `Headers` input, therefore testing that method is included in the unit test:

```php
$message = new Message(['headers' => new Headers(), 'content' => (string)$body]);

// Mime\Decode::splitMessage calls toString on headers object which creates invalid result:
if ($message instanceof Headers) {
    $message = $message->toString();
}
```

---

Originally posted by @glensc at https://github.com/zendframework/zend-mail/pull/230

weierophinney:
@weierophinney, @Xerkus can you take care of this?

---

Originally posted by @glensc at https://github.com/zendframework/zend-mail/pull/230#issuecomment-508890560

weierophinney:
this is a regression from #147 as it added `;` separator support that's why (i believe) this exception is caught.

however, there should not be an assumption that only zend-mail is used for reading writing mail messages.

---

Originally posted by @glensc at https://github.com/zendframework/zend-mail/pull/230#issuecomment-508890658

PR #93, opened 2020-06-27
Fix AddressList toString method to quote semicolon

glensc:
This is port of https://github.com/zendframework/zend-mail/pull/230 fixing regression from https://github.com/zendframework/zend-mail/pull/147

Fixes https://github.com/laminas/laminas-mail/issues/14

----

Certain input of `AddressList` headers, cannot be converted to string and back to header object because of incorrect quoting:

```
From: "Foo;" <root@example.com>
```
gets incorrectly converted as

```
From: Foo; <root@example.com>
```

but `;` is address separator, so it needs to be quoted:

```
From: "Foo;" <root@example.com>
```

The problem I discovered internally when using `Storage\Message` with `Headers` input, therefore testing that method is included in the unit test:

```php
$message = new Message(['headers' => new Headers(), 'content' => (string)$body]);

// Mime\Decode::splitMessage calls toString on headers object which creates invalid result:
if ($message instanceof Headers) {
    $message = $message->toString();
}
```

### DIFF
--- a/src/Header/AbstractAddressList.php
+++ b/src/Header/AbstractAddressList.php
@@ -135,7 +135,9 @@ abstract class AbstractAddressList implements HeaderInterface
             $email = $address->getEmail();
             $name  = $address->getName();
 
-            if (! empty($name) && false !== strstr($name, ',')) {
+            // quote $name if value requires so
+            if (! empty($name) && (false !== strpos($name, ',') || false !== strpos($name, ';'))) {
+                // FIXME: what if name contains double quote?
                 $name = sprintf('"%s"', $name);
             }
 
@@ -241,7 +243,7 @@ abstract class AbstractAddressList implements HeaderInterface
      * Supposed to be private, protected as a workaround for PHP bug 68194
      *
      * @param string $value
-     * @return void
+     * @return string
      */
     protected static function stripComments($value)
     {
--- a/test/Storage/MessageTest.php
+++ b/test/Storage/MessageTest.php
@@ -10,6 +10,7 @@ namespace LaminasTest\Mail\Storage;
 
 use Exception as GeneralException;
 use Laminas\Mail\Exception as MailException;
+use Laminas\Mail\Headers;
 use Laminas\Mail\Storage;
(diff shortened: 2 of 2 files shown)


glensc:
NOTE: This is needed to switch to laminas main in the project I maintain: https://github.com/eventum/eventum/pull/876

michalbundyra:
@glensc thanks for your contribution. Before we can accept the PR with need DCO to be valid on PRs in Laminas org. Please see https://github.com/laminas/laminas-mail/pull/93/checks?check_run_id=813642578 for more details. Thanks!

glensc:
@michalbundyra sure, those were fixup commits that I intended to flatten once Travis passed

src/Header/AbstractAddressList.php
```diff
@@ -135,7 +135,9 @@ public function getFieldValue($format = HeaderInterface::FORMAT_RAW)
             $email = $address->getEmail();
             $name  = $address->getName();
 
-            if (! empty($name) && false !== strstr($name, ',')) {
+            // quote $name if value requires so
+            if (! empty($name) && (false !== strpos($name, ',') || false !== strpos($name, ';'))) {
+                // FIXME: what if name contains double quote?
```
michalbundyra, line 140: @glensc would be great, if you can provide another PR with a proper test case for that, thanks!

glensc: @michalbundyra what you mean proper? what's wrong with the provided one? please be specific what you mean, test case for what exactly?

michalbundyra: @glensc sorry, I mean test with name which contains double quote, so separate issue. 

As I understand this is about `;` in the name only.

glensc: perhaps I remove this "FIXME" comment then it being out of scope? as the changeset does not introduce that as a new problem, the problem was there already before, I just noticed that and left a note.

besides I don't know how it should be encoded and at this point and I'm not interested in researching on, this PR has been waiting to be accepted by upstream way too long to get sidetracked with this.

PR #95, opened 2020-07-23
Fixes #39 Fixes #40 remove laminas-loader dependency

### LINKED ISSUE #39: Drop dependency on zendframework/zend-loader
This removes the dependency on the zendframework/zend-load package, as suggested in #185 

The `HeaderLoader` class has been removed and replaced by a simple class map in the `Headers` class.

However, I have also removed the `getPluginClassLoader` and `setPluginClassLoader` methods. Since they are public, this might be a BC break. What do you think?

---

Originally posted by @acelaya at https://github.com/zendframework/zend-mail/pull/186

### LINKED ISSUE #40: Consider removing dependency on zend-loader
I have seen that this package requires `zendframework/zend-loader`, just for one class, the `Zend\Mail\Header\HeaderLoader`.

However, this class basically defines a map to resolve header names to class names.

Would you be open to replace this by a simple array map in the `Zend\Mail\Headers` class? This would let us drop the requirement on `zendframework/zend-loader` (assuming I'm not missing anything...).

I could do it if you want.

---

Originally posted by @acelaya at https://github.com/zendframework/zend-mail/issues/185

samsonasik:
|    Q          |   A
|-------------- | ------
| New Feature | Yes
| BC Break      |  No

### Description

Re-create and modify https://github.com/zendframework/zend-mail/pull/186 by @acelaya

Fixes #39
Fixes #40 

Original PR description:

```
This removes the dependency on the zendframework/zend-load package, as suggested in #185

The HeaderLoader class has been removed and replaced by a simple class map in the Headers class.

However, I have also removed the getPluginClassLoader and setPluginClassLoader methods.
```

## Actual changes

- Bumps minimum supported PHP version to 7.1
- Introduces `Laminas\Mail\Header\HeaderLocatorInterface`, defining how to map header names to classes.
- Introduces `Laminas\Mail\Header\HeaderLocator` as a default implementation of that interface.
- Modifies `Laminas\Mail\Headers` as follows:
  - Marks `setPluginClassLoader()` and `getPluginClassLoader()` as deprecated.
  - Adds a `$headerLocator` property and accompanying `setHeaderLocator()` and `getHeaderLocator()` methods.
  - Adds a private method, `resolveHeaderClass()` that:
    - if a non-null `$pluginClassLoader` property is available, uses that to locate the header class.
    - otherwise, uses the return value of `getHeaderLocator()` to locate the header class.

This approach provides a forwards-compatible approach to removing laminas-loader in version 3.0.


src/Headers.php
```diff
@@ -528,4 +502,17 @@ protected function normalizeFieldName($fieldName)
...
+     * @param string $key
+     * @return string
+     */
+    private function resolveHeaderClass($key)
+    {
+        if ($this->headerLoader === null) {
+            $this->headerLoader = new Header\HeaderLoader();
+        }
+
+        return $this->headerLoader->get($key, Header\GenericHeader::class);
```
samsonasik, line 516: @acelaya this seems cause an issue when get header by key string to use correct header class.

samsonasik: It seems I miss-copy plugin that should be new headerClassMap property.

samsonasik:
travis build green 🎉

glensc:
@samsonasik please copy the description from original PR, so wouldn't have to click dozen of links to get needed information. for example, I'd like to know why this is BC break.

samsonasik:
@glensc updated, thanks ;)

glensc:
@samsonasik info why this is BC break still missing.

samsonasik:
The removing of public method is BC break, that is in the description.

glensc:
@samsonasik write it so then

samsonasik:
that is in the description that I copied

glensc:
@samsonasik seems you didn't get my message saying that copy text from original issue, so wouldn't have to click dozen of links to get the needed information. for example, I'd like to know **why** this is BC break. now that info is spread around in comments, not in compact form for current and future readers can find in one place.

src/Headers.php
```diff
@@ -118,31 +117,6 @@ public static function fromString($string, $EOL = self::EOL)
...
-     * @return PluginClassLocator
-     */
-    public function getPluginClassLoader()
-    {
-        if ($this->pluginClassLoader === null) {
-            $this->pluginClassLoader = new Header\HeaderLoader();
-        }
-        return $this->pluginClassLoader;
-    }
-
```
weierophinney, line 145: This part is the BC break; users calling those methods now will have errors occur.

There are two options:

- Bump to a new major version. This is not terribly enticing, as new major versions often take time to get traction, and we'd likely want to delay release until there are other BC breaks in place that we want to achieve.
- Keep the methods, but change them to indicate deprecation, as well as to modify how they work slightly.

If we went the second route:

- We can remove the parameter typehint on `setPluginClassLoader()` (type widening is allowed for parameters), and internally check for either a `HeaderLoader` instance OR a laminas-loader `PluginClassLocator`. In the second case, we'd raise a deprecation notice.
- We'd change the return typehint on `getPluginClassLoader()` to `HeaderLoader|PluginClassLocator`, and raise a deprecation notice if it is called.

We'd need to ensure that the same property name is used by both these methods and `resolveHeaderClass()`.

samsonasik: @weierophinney  I've updated keep setPluginClassLoader() and getPluginClassLoader() with E_USER_DEPRECATION mark and keep property pluginClassLoader

weierophinney:
@samsonasik I pushed additional changes to your branch. Basically:

- Added accessors for the header loader. This sort of extensibility is necessary so that users can accommodate custom header types. By providing new accessors, we detach them from the naming of the old implementation.

- Modified resolveHeaderClass to call on the new `getHeaderLoader()` to do lazy loading.

- Reverted a removed test for multi-part header lines.

- Added tests to ensure that the deprecation notices are being emitted.

- Added tests for the new accessors.

I think it's ready to ship now!

michalbundyra:
@weierophinney 

> Removing the extends PluginClassLoader from HeaderLoader without implementing the public methods

just removing `extends PluginClassLoader` is a BC Break.

What we can do (something I do not really like) is create class `PluginClassLoader` from `Laminas\Loader` namespace here, mark as deprecated (only if laminas-loader is not installed) and use it instead.

glensc:
@michalbundyra, you mean something like this on top of `HeaderLoader.php` file:

```php
use Laminas\Loader\PluginClassLoader;

if (!class_exists(PluginClassLoader::class)) {
   class PluginClassLoader {}
}
```

might need to scope the namespace (needs testing):

```php
namespace Laminas\Loader {
  if (!class_exists(PluginClassLoader::class)) {
     class PluginClassLoader {}
  }
}

```

weierophinney:
> you mean something like this on top of `HeaderLoader.php` file

No - it would be a separate class file. (Our CS doesn't allow for multiple namespaces and/or class definitions in the same file.)

glensc:
@weierophinney how do you trigger autoloader then? via autoload.files?

- https://getcomposer.org/doc/04-schema.md#files

weierophinney:
@glensc —

> I think you should do Symfony-style silent deprecation messages

That's a cool technique; I'd not seen using `@` with `trigger_error` before.  I'll update the code accordingly.

> how do you trigger autoloader then? via autoload.files?

Yes. Generally, we've created an `autoload.php` file that checks to see if the class exists, and, if not, it then loads the file defining it. In this case, we'd need to provide both the interface and the implementation.

However, I don't think it's going to happen right now. More below.

---

@samsonasik and @michalbundyra —

I've been thinking on this a ton, and there's no clean way to do this in a BC manner. However, we can make a change that provides forwards compatibility and formally deprecates usage of laminas-loader, so that a v3 can remove it.

Here's how.

- We leave `HeaderLoader` unchanged.

- We create a new interface, `HeaderLocatorInterface`, that defines a `resolveHeaderClass(string $headerName): string` method.

- We create a new class, `HeaderLocator`, that implements this new interface and provides essentiallly the same API as you have proposed for `HeaderLoader`, to make it easy to add custom `HeaderInterface` implementations to the mix.

- The method `setHeaderLocator()` is modified to accept the new interface, and the `getHeaderLocator()` method is modified to return the new interface, lazy-loading the new `HeaderLocator` implementation when none is provided. They each work with a new `$headerLocator` property.

- Internally, `resolveHeaderClass()`:
  - Checks if `$pluginClassLoader` is `null`. If not, it uses that to load the requested header.
  - If `$pluginClassLoader` is `null`, it calls `getHeaderLocator()` to retrieve the `HeaderLocator` instance, and then the value of calling its `get()` method.

We then document in both the CHANGELOG and the documentation the new interface, implementation, and default behavior, so that users know that they need to start modifying their code if they relied on the `PluginClassLocator` implementation.

Thoughts?

src/Headers.php
```diff
@@ -119,30 +122,73 @@ public static function fromString($string, $EOL = self::EOL)
     }
 
     /**
-     * Set an alternate implementation for the PluginClassLoader
+     * Set an alternate implementation for loading header classes.
      *
-     * @param  PluginClassLocator $pluginClassLoader
+     * @deprecated since 2.12.0
+     * @todo Remove for version 3.0.0
+     * @param  Header\HeaderLoader|\Laminas\Loader\PluginClassLocator $pluginClassLoader
```
samsonasik, line 129: `Laminas\Loader\PluginClassLocator` can be changed with `PluginClassLocator` as the class re-added to use statement.

samsonasik:
@weierophinney I think that `HeaderLocator` will be ok, but that means `laminas-loader` will only in `require-dev`? or still in `require` ?

weierophinney:
> I think that `HeaderLocator` will be ok, but that means `laminas-loader` will only in `require-dev`? or still in `require` ?

Still in `require` until we can remove it in 3.0. Essentially, I'm proposing we change the scope of this PR to make a forwards-compatible change so we can deprecate the usage of laminas-loader, since we have (a) a class that extends a class in that package, and (b) methods that are tied to an interface from that package, and (c) the latter are the default way to manipulate what custom headers can be handled by end-users. The PR would offer another way of doing this that is not tied to an external package, allowing us to **deprecate** those features, and then **remove them in 3.0**.

I'm willing to do the work, if you'd like.

samsonasik:
@weierophinney ok, I will let you continue. Thank you

glensc:
you can use composer "replace" trick to remove "laminas-loader" from your installed vendors:
- https://github.com/glensc/eventum/blob/d94105835b32af5750592cb70cbcb8fa3c9f9252/composer.json#L90-L91

but this solution is suitable only for applications. don't do this in libraries or your users will hate you.

weierophinney:
@michalbundyra and/or @samsonasik — changes are in place. I've updated the description to better detail what has finally happened, and detailed it in the changelog as well. Tests are green. :)

src/Headers.php
```diff
@@ -119,30 +128,73 @@ public static function fromString($string, $EOL = self::EOL)
...
-     * @param  PluginClassLocator $pluginClassLoader
-     * @return Headers
+     * @deprecated since 2.12.0
+     * @todo Remove for version 3.0.0
+     * @return $this
      */
     public function setPluginClassLoader(PluginClassLocator $pluginClassLoader)
     {
+        // Silenced; can be caught in custom error handlers.
+        @trigger_error(sprintf(
```
weierophinney, line 140: As @glensc pointed out, this is how Symfony does deprecation notices. They'll get caught in error handlers, but not if unhandled. This prevents breakage in most use cases, but if you log errors, you'll catch them in your logs.

src/Headers.php
```diff
@@ -270,9 +322,20 @@ public function addHeader(Header\HeaderInterface $header)
...
     public function removeHeader($instanceOrFieldName)
     {
+        if (! $instanceOrFieldName instanceof Header\HeaderInterface && ! is_string($instanceOrFieldName)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s requires a string or %s instance; received %s',
+                __METHOD__,
+                Header\HeaderInterface::class,
+                is_object($instanceOrFieldName) ? get_class($instanceOrFieldName) : gettype($instanceOrFieldName)
+            ));
+        }
```
weierophinney, line 332: This bit was interesting. Because I turned on strict types in this file, I discovered a test that was passing `null` to this method, which was incorrect. As such, I added this guard (and updated the test!).

glensc:
@weierophinney can you add a summary to users what they should do to stop using zend-loader?

just upgrade and add composer "replace" to loader package?
- https://github.com/laminas/laminas-mail/pull/95#issuecomment-665943569

assuming just used defaults previously

weierophinney:
@glensc By default, if you are NOT calling `getPluginClassLoader()` or `setPluginClassLoader()`, `Headers` will no longer use the laminas-loader functionality. This is spelled out in the changelog:

> **The value of `getHeaderLocator()` will now be used as the default mechanism for resolving header names to header classes.**

Beyond that, if you're using the laminas-mvc, you cannot really replace the laminas-loader package, as it's used by the MVC (though the usage is largely deprecated). As such, I haven't detailed how to do that; it's a niche use case. But what it _does_ do is open the door for us to remove it from this component in v3, and, if v3 is not released before releases of MVC components that do not use it, we can replace it in the laminas-mvc-skeleton package.

PR #96, opened 2020-07-24
Fix a space interference

hundan2020:
sometimes may get parameter like " name", the normal way might be "name",

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Tell us about why this change is necessary:
- Are you fixing a bug or providing a failing unit test to demonstrate a bug?

fix a bug

  - How do you reproduce it?

```
------=_NextPart_5F183DBC_1015AD90_719B01C0
Content-Type: application/octet-stream;
	charset="gb18030";
	name="=?gb18030?B?a2sgMjAyMC0wNy0yMiAyMS0wMC01OS5tcDQ=?="
Content-Disposition: attachment; filename="=?gb18030?B?a2sgMjAyMC0wNy0yMiAyMS0wMC01OS5tcDQ=?="
Content-Transfer-Encoding: base64

```

```
$m->current()->getHeader('Content-Type')->getParameter("name")
```

  - What did you expect to happen?

```
kk 2020-07-22 21-00-59.mp4
```

  - What actually happened?

```
null
```

  - TARGET THE master BRANCH

- Are you adding documentation? **no**
  - TARGET THE master BRANCH

- Are you providing a QA improvement (additional tests, CS fixes, etc.) that
  does not change behavior? **no**
  - Explain why the changes are necessary **to get the correct parameter, easy to use**
  - TARGET THE master BRANCH

- Are you fixing a BC Break?  **no**
  - How do you reproduce it?
  - What was the previous behavior?
  - What is the current behavior?
  - TARGET THE master BRANCH

- Are you adding something the library currently does not support? **no**
  - Why should it be added?
  - What will it enable?
  - How will the code be used?
  - TARGET THE develop BRANCH

- Are you refactoring code?  **no**
  - Why do you feel the refactor is necessary?
  - What types of refactoring are you doing?
  - TARGET THE develop BRANCH

### DIFF
--- a/src/Header/ContentType.php
+++ b/src/Header/ContentType.php
@@ -155,7 +155,7 @@ class ContentType implements UnstructuredInterface
             );
         }
 
-        $this->parameters[$name] = $value;
+        $this->parameters[trim($name)] = $value;
         return $this;
     }
 


samsonasik:
I created PR #97 for its alternative, with include unit test.

PR #97, opened 2020-07-24
Fixes #23 ensure trim() on ContentType::addParameter()

### LINKED ISSUE #23: Content-Type paramater does not trim paramater name
Hello, 

I am getting filenames for attachments using the getParameter() on the content-type header.  I noticed some email parts that were supposed to be attachments did not have a filename.  Inspecting the source header, I could see that there was in fact a filename set.  Further inspection showed that the parameter name was being added with a leading space.  I would expect that leading spaces would be trimmed.

Thanks,

Cody

### Code to reproduce the issue
```php
use Zend\Mail\Header\ContentType;
$ct = ContentType::fromString('Content-Type: text/plain; charset=utf-8; name="logfile.log";');
printf('r1 "%s"'.PHP_EOL, $ct->getParameter('name'));
printf('r2 "%s"'.PHP_EOL, $ct->getParameter(' name'));
var_dump($ct);
```
### Actual results
```
r1 ""
r2 "logfile.log"
object(Zend\Mail\Header\ContentType)#3 (3) {
  ["type":protected]=>
  string(10) "text/plain"
  ["encoding":protected]=>
  string(5) "ASCII"
  ["parameters":protected]=>
  array(2) {
    ["charset"]=>
    string(5) "utf-8"
    [" name"]=>
    string(11) "logfile.log"
  }
}
```
### My Fix
I would not presume to understand the best place to resolve this, but this hack solved my problem for the time being

```
$ diff ContentType.php ContentType.php.new
55c55
<                 $header->addParameter($values[$i], $value);
---
>                 $header->addParameter(trim($values[$i]), $value);
```

---

Originally posted by @cody121 at https://github.com/zendframework/zend-mail/issues/220

samsonasik:

|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Fixes #23 ensure `trim()` on `ContentType::addParameter()` so `ContentType::getParameter("name")` will always get correct result.

### DIFF
--- a/src/Header/ContentType.php
+++ b/src/Header/ContentType.php
@@ -143,7 +143,7 @@ class ContentType implements UnstructuredInterface
      */
     public function addParameter($name, $value)
     {
-        $name  = strtolower($name);
+        $name  = trim(strtolower($name));
         $value = (string) $value;
 
         if (! HeaderValue::isValid($name)) {
--- a/test/Header/ContentTypeTest.php
+++ b/test/Header/ContentTypeTest.php
@@ -222,6 +222,12 @@ class ContentTypeTest extends TestCase
         $this->assertSame('top', $header->getParameter('level'));
     }
 
+    public function testGetParameterWithSpaceTrimmed()
+    {
+        $header = ContentType::fromString('content-type: text/plain; level=top; name="logfile.log";');
+        $this->assertSame('logfile.log', $header->getParameter('name'));
+    }
+
     public function testGetParameterNotExists()
     {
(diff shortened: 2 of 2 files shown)


samsonasik:
travis build green :tada:

glensc:
IMHO this is fixed in the wrong place. the trim should be in a call to addParameter not in addParameter itself.

michalbundyra:
@glensc Please open separate PR and link to this issue. As it is closed already can be easily missed. Thanks!

Issue #36, opened 2020-01-15
remove Null Transport


michalbundyra:
since composer.json master branch is only support php:^7.1

---

Originally posted by @samsonasik at https://github.com/zendframework/zend-mail/pull/195

michalbundyra:
Please rebase this against develop and change the target branch. Since it's a BC break, it needs to target a new major version.

---

Originally posted by @weierophinney at https://github.com/zendframework/zend-mail/pull/195#issuecomment-369737328

michalbundyra:
@weierophinney I've rebased against develop and change target to develop branch.

---

Originally posted by @samsonasik at https://github.com/zendframework/zend-mail/pull/195#issuecomment-369738376

michalbundyra:
i suppose it addresses problem that `null` is no longer allowed as class names in newer php versions? 

and explanation why it's *removed* not just renamed?


also, fix spelling: `Tranport` -> `Transport`, 

---

Originally posted by @glensc at https://github.com/zendframework/zend-mail/pull/195#issuecomment-370810847

michalbundyra:
the `null` transport was mapped to `inMemory`, so, it is same to use `inMemory`. Thank you, I've updated PR title  `Tranport -> Transport,`

---

Originally posted by @samsonasik at https://github.com/zendframework/zend-mail/pull/195#issuecomment-370942877

michalbundyra:
This repository has been moved to laminas/laminas-mail. If you feel that this patch is still relevant, please re-open against that repository, and reference this issue. To re-open, we suggest the following workflow:
 - Squash all commits in your branch (`git rebase -i origin/{branch}`)
 - Make a note of all changed files (`git diff --name-only origin/{branch}...HEAD
 - Run the laminas/laminas-migration tool on the code.
 - Clone laminas/laminas-mail to another directory.
 - Copy the files from the second bullet point to the clone of laminas/laminas-mail.
 - In your clone of laminas/laminas-mail, commit the files, push to your fork, and open the new PR.
 We will be providing tooling via laminas/laminas-migration soon to help  automate the process.

---

Originally posted by @weierophinney at https://github.com/zendframework/zend-mail/pull/195#issuecomment-569994627

samsonasik:
I've created PR https://github.com/laminas/laminas-mail/pull/98 for it.

PR #98, opened 2020-07-24
remove null transport

samsonasik:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Re-create https://github.com/zendframework/zend-mail/pull/195 . Null is reserved keyword in php 7+. It is for next major version when minimum php version upped to php 7+. Fixes https://github.com/laminas/laminas-mail/issues/36


weierophinney:
> It is for next major version when minimum php version upped to php 7+.

Well, we can bump the minimum version to 7 anytime. The problem is that this class is completely unusable when we do. As such, I'm going to wrap it in the 2.12.0 changes.

test/Transport/FactoryTest.php
```diff
@@ -102,15 +98,12 @@ public function typeAliasProvider()
     {
         return [
             ['file', File::class],
-            ['null', InMemory::class],
```
weierophinney, line 105: We can actually keep these references to keep BC for end-users.

src/Transport/Factory.php
```diff
@@ -20,7 +20,6 @@ abstract class Factory
         'file'      => File::class,
         'inmemory'  => InMemory::class,
         'memory'    => InMemory::class,
-        'null'      => InMemory::class,
```
weierophinney, line 23: We can keep this, as it's a string; removing it would be a BC break.

PR #103, opened 2020-08-05
Fixes #19 Fixes #72 duplicate subject header on windows

### LINKED ISSUE #19: Bugfix/duplicat subject header
This fixes the issue described in https://github.com/zendframework/zend-mail/issues/22

---

Originally posted by @arueckauer at https://github.com/zendframework/zend-mail/pull/225

### LINKED ISSUE #72: Subject header is duplicated
**Summary**
On Windows with PHP 7, when sending an mail message with the Sendmail transport, the subject header is duplicated.

**Steps to reproduce**
To reproduce the bug the sample from here can be used:
http://framework.zend.com/manual/current/en/modules/zend.mail.introduction.html

The resulting mail headers look something like this:
Subject: Test Subject
Date: Wed, 29 Jul 2015 17:33:17 +0000
From: =?UTF-8?Q?test?= test@test.com
To: test@test.com
Subject: Test Subject

**Cause**
In the Sendmail transport all headers (including the Subject) and the subject itself are passed to the php mail function:
$result = mail($to, $subject, $message, $headers);
The mail function then appends to the $subject to the $headers. I do not think the php mail function should handle this case because it doesn't know what subject the caller intended to use. However I believe php mail should at least issue a warning (TBD: file enhancement at https://bugs.php.net/).

**Solution**
Pass the headers without the Subject header to the php mail function. This is already done if the os is not Windows in the prepareHeaders. I think it should be done for Windows as well.

---

Originally posted by @matbech at https://github.com/zendframework/zend-mail/issues/22

samsonasik:

|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Re-create https://github.com/zendframework/zend-mail/pull/225 by @arueckauer Fixes #19 #72 . Original issue description reported by @matbech : 

#### Summary

On Windows with PHP 7, when sending an mail message with the Sendmail transport, the subject header is duplicated.

#### Steps to reproduce

To reproduce the bug the sample from here can be used:
http://framework.zend.com/manual/current/en/modules/zend.mail.introduction.html

The resulting mail headers look something like this:

```text
Subject: Test Subject
Date: Wed, 29 Jul 2015 17:33:17 +0000
From: =?UTF-8?Q?test?= test@test.com
To: test@test.com
Subject: Test Subject
```

#### Cause

In the Sendmail transport all headers (including the Subject) and the subject itself are passed to the php mail function:

```php
$result = mail($to, $subject, $message, $headers);
```

The mail function then appends to the `$subject` to the `$headers`. I do not think the php mail function should handle this case because it doesn't know what subject the caller intended to use. However I believe php mail should at least issue a warning (TBD: file enhancement at https://bugs.php.net/).

### DIFF
--- a/test/Transport/SendmailTest.php
+++ b/test/Transport/SendmailTest.php
@@ -73,7 +73,7 @@ class SendmailTest extends TestCase
         return $message;
     }
 
-    public function isWindows()
+    private function isWindows()
     {
         return $this->operating_system === 'WIN';
     }
@@ -263,12 +263,10 @@ class SendmailTest extends TestCase
     }
 
     /**
-     * @see @see https://github.com/laminas/laminas-mail/issues/19
+     * @see https://github.com/laminas/laminas-mail/issues/19
      */
     public function testHeadersToAndSubjectAreNotDuplicated()
     {
-        $lineBreak = $this->isWindows() ? "\r\n" : "\n";
-
         $message = new Message();
         $message
             ->addTo('matthew@example.org')
@@ -282,6 +280,6 @@ class SendmailTest extends TestCase
         $this->assertEquals('Greetings and Salutations!', $this->subject);
 
         $this->assertNotRegExp('/^To: matthew\@example\.org$/m', $this->additional_headers);
-        $this->assertNotRegExp('/^Subject: Greetings and Salutations\!$/m', $this->additional_headers);
+        $this->assertNotRegExp('/^Subject: Greetings and Salutations!$/m', $this->additional_headers);
     }
 }
--- a/test/Transport/SendmailTest.php
+++ b/test/Transport/SendmailTest.php
@@ -281,13 +281,7 @@ class SendmailTest extends TestCase
         $this->assertEquals('matthew@example.org', $this->to);
         $this->assertEquals('Greetings and Salutations!', $this->subject);
 
-        $this->assertNotContains(
-            'To: matthew@example.org, matthew@example.org' . $lineBreak,
-            $this->additional_headers
-        );
-        $this->assertNotContains(
-            'Subject: Greetings and Salutations!, Greetings and Salutations!' . $lineBreak,
-            $this->additional_headers
-        );
+        $this->assertNotRegExp('/^To: matthew\@example\.org$/m', $this->additional_headers);
+        $this->assertNotRegExp('/^Subject: Greetings and Salutations\!$/m', $this->additional_headers);
     }
 }
--- a/test/Transport/SendmailTest.php
+++ b/test/Transport/SendmailTest.php
@@ -28,6 +28,7 @@ class SendmailTest extends TestCase
     public $message;
     public $additional_headers;
     public $additional_parameters;
+    public $operating_system;
 
     public function setUp()
     {
@@ -72,9 +73,14 @@ class SendmailTest extends TestCase
         return $message;
     }
 
+    public function isWindows()
+    {
+        return $this->operating_system === 'WIN';
+    }
+
     public function testReceivesMailArtifactsOnUnixSystems()
     {
-        if ($this->operating_system == 'WIN') {
+        if ($this->isWindows()) {
             $this->markTestSkipped('This test is *nix-specific');
         }
 
@@ -96,7 +102,7 @@ class SendmailTest extends TestCase
 
     public function testReceivesMailArtifactsOnWindowsSystems()
     {
-        if ($this->operating_system != 'WIN') {
+        if (! $this->isWindows()) {
             $this->markTestSkipped('This test is Windows-specific');
         }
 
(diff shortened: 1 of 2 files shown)


test/Transport/SendmailTest.php
```diff
@@ -255,4 +261,33 @@ public function testDoNotAllowMessageWithoutToAndCcAndBccHeaders()
...
+        $this->assertEquals('Greetings and Salutations!', $this->subject);
+
+        $this->assertNotContains(
+            'To: matthew@example.org, matthew@example.org' . $lineBreak,
+            $this->additional_headers
+        );
+        $this->assertNotContains(
+            'Subject: Greetings and Salutations!, Greetings and Salutations!' . $lineBreak,
+            $this->additional_headers
+        );
```
weierophinney: This is not the scenario indicated in the original report.

The original report was indicating that the **entire** subject header was being emitted _twice_ in the headers, not that the subject header **value** was being duplicated. The reporter observed the following output:

```text
Subject: Greetings and Salutations!
To: matthew@example.org
From: ralph@example.org
Subject: Greetings and Salutations!
```

Further, their assertion was that this was due to the fact that `mail()` was being called with a subject, and the underlying library emitting a subject header from that value _as well as_ emitting the one in the `$additional_headers` argument in that function.

Change the above to:

```php
$this->assertNotRegExp('/^Subject: Greetings and Salutations\!$/m', $this->additional_headers);
```

The "m" PCRE modifier will test each _line_ of the text against the regexp, allowing you to use normal start-of-line/end-of-line anchors - and you can thus ignore linebreaks, making the comparison simpler!

weierophinney (changes requested):

test/Transport/SendmailTest.php
weierophinney: Also, technically the `To` header will not be present in the additional headers, as we strip it to pass as a `mail()` argument; as such, you'll also have:

```php
$this->assertNotRegExp('/^To: matthew\@example\.org$/m', $this->additional_headers);
```

samsonasik: @weierophinney updated.

test/Transport/SendmailTest.php
```diff
@@ -255,4 +261,27 @@ public function testDoNotAllowMessageWithoutToAndCcAndBccHeaders()
...
+            ->setSubject('Greetings and Salutations!')
+            ->setBody("Sorry, I'm going to be late today!");
+
+        $this->transport->send($message);
+
+        $this->assertEquals('matthew@example.org', $this->to);
+        $this->assertEquals('Greetings and Salutations!', $this->subject);
+
+        $this->assertNotRegExp('/^To: matthew\@example\.org$/m', $this->additional_headers);
+        $this->assertNotRegExp('/^Subject: Greetings and Salutations\!$/m', $this->additional_headers);
```
glensc: `!` does not need to be escaped\\!

`
        $this->assertNotRegExp('/^Subject: Greetings and Salutations!$/m', $this->additional_headers);
`

test/Transport/SendmailTest.php
```diff
@@ -72,9 +73,14 @@ public function getMessage()
         return $message;
     }
 
+    public function isWindows()
```
glensc: `    private function isWindows()
`

do not export as public if not needed

test/Transport/SendmailTest.php
```diff
@@ -255,4 +261,27 @@ public function testDoNotAllowMessageWithoutToAndCcAndBccHeaders()
         $this->expectException(RuntimeException::class);
         $this->transport->send($message);
     }
+
+    /**
+     * @see @see https://github.com/laminas/laminas-mail/issues/19
+     */
+    public function testHeadersToAndSubjectAreNotDuplicated()
+    {
+        $lineBreak = $this->isWindows() ? "\r\n" : "\n";
```
glensc: drop unused `$lineBreak` variable.

test/Transport/SendmailTest.php
```diff
@@ -255,4 +261,27 @@ public function testDoNotAllowMessageWithoutToAndCcAndBccHeaders()
         $this->expectException(RuntimeException::class);
         $this->transport->send($message);
     }
+
+    /**
+     * @see @see https://github.com/laminas/laminas-mail/issues/19
```
glensc: remove duplicate `@see`

samsonasik: fixed.

test/Transport/SendmailTest.php
samsonasik: fixed.

test/Transport/SendmailTest.php
samsonasik: fixed.

test/Transport/SendmailTest.php
samsonasik: fixed.

weierophinney (approved):

PR #104, opened 2020-08-05
Fixes #17 : Fixes when set -f option, no set -f option automatically

### LINKED ISSUE #17: When set -f option, no set -f option automatically
`Zend\Mail\Transport\SendMail` always append envelope-from parameter.  
But I want to set any envelope-from.  
In this PR, only when not set option `-f`, set `-f` option automatically.
This means if I set `-ffoo@example.com` or `-bs -ffoo@example.com`, `Zend\Mail\Transport\Sendmail` doesn't set `-f` option.

---

Originally posted by @tzmfreedom at https://github.com/zendframework/zend-mail/pull/227

samsonasik:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Re-create https://github.com/zendframework/zend-mail/pull/227 by @tzmfreedom Fixes #17 .  Original PR description : 

```
Zend\Mail\Transport\SendMail always append envelope-from parameter.
But I want to set any envelope-from.
In this PR, only when not set option -f, set -f option automatically.
This means if I set -ffoo@example.com or -bs -ffoo@example.com, Zend\Mail\Transport\Sendmail doesn't set -f option.
```

### DIFF
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -255,7 +255,7 @@ class Sendmail implements TransportInterface
         }
 
         $parameters = (string) $this->parameters;
-        if (preg_match('/^\-f.+$/', $parameters) || strpos($parameters, ' -f') !== false) {
+        if (preg_match('/(^| )\-f.+/', $parameters)) {
             return $parameters;
         }
 
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -255,6 +255,9 @@ class Sendmail implements TransportInterface
         }
 
         $parameters = (string) $this->parameters;
+        if (preg_match('/^\-f.+$/', $parameters) || strpos($parameters, ' -f') !== false) {
+            return $parameters;
+        }
 
         $sender = $message->getSender();
         if ($sender instanceof AddressInterface) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ All notable changes to this project will be documented in this file, in reverse 
 
 - [#103](https://github.com/laminas/laminas-mail/pull/103) fixes issues on Windows whereby the "Subject" and "To" headers were duplicated.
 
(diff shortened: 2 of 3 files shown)


src/Transport/Sendmail.php
```diff
@@ -260,6 +260,9 @@ protected function prepareParameters(Mail\Message $message)
         }
 
         $parameters = (string) $this->parameters;
+        if (preg_match('/^\-f.+$/', $parameters) || strpos($parameters, ' -f') !== false) {
```
weierophinney:
```suggestion
        if (strpos($parameters, '-f') !== false) {
```

This accomplishes the same thing, with one less operation.

weierophinney (changes requested):
I've made one suggestion to how you check for the `-f` flag.

Additionally, we need tests demonstrating that the `setParameters()` method uses the parameter passed to the transport and NOT the sender and/or from address when it is provided; currently, you only demonstrate it is used when provided, and not used when not present, but not the additional case where it is not provided and falls through to the sender or from address.

src/Transport/Sendmail.php
michalbundyra: @weierophinney it doesn't seem right, imo

it's not the same, previously we accept `-f` at the beginning of when it is after a space.

I can imagine case `my-fancy-param` where `-f` is no at the beginning nor after a space.

weierophinney: Then do it as follows:

```php
if (preg_match('/(^| )\-f.+/', $parameters)) {
```

It doesn't make sense to have two conditions when it's essentially the same condition ("is there a -f switch provided?"). The above regex does the "at the beginning" as well as "following a space" conditions.

michalbundyra: that's better ;-) there is also redundant `\` before `-` in the regexp

samsonasik: @weierophinney updated.

weierophinney (approved):
I've combined the two unit tests into a single one backed by a data provider. I think it's ready to :ship: once tests have run.

PR #110, opened 2020-08-24
Allow code to access the last response string.

Piskvor:
This does make the class stateful - OTOH that reflects the nature of an IMAP session.
- requestAndResponse() provides very rudimentary response for debugging negative responses
- it is at times useful to see the exact response of the server
- "NO [AUTHENTICATIONFAILED] Authentication failed." is far more informative than a FALSE
- also, there's currently no way to distinguish between different error states
- e.g. the above and "NO [INSUFFICIENTGROVEL] Zot!"

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

- Are you adding something the library currently does not support?
  - Why should it be added? Users of this library are likely to be of technical nature, and might need the information for troubleshooting
  - What will it enable? A more verbose error reporting, if requested by the user.
  - How will the code be used? `$imapProtocol->login() or log_error($imapProtocol->getLastResponse());`
  - TARGET THE develop BRANCH

### DIFF
--- a/src/Protocol/Imap.php
+++ b/src/Protocol/Imap.php
@@ -29,6 +29,12 @@ class Imap
     protected $tagCount = 0;
 
     /**
+     * Last line returned from server
+     * @var string|null
+     */
+    protected $lastResponseLine;
+
+    /**
      * Public constructor
      *
      * @param  string       $host           hostname or IP address of IMAP server, if given connect() is called
@@ -128,8 +134,8 @@ class Imap
      */
     protected function assumedNextLine($start)
     {
-        $line = $this->nextLine();
-        return strpos($line, $start) === 0;
+        $this->lastResponseLine = $this->nextLine();
+        return strpos($this->lastResponseLine, $start) === 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Protocol/Imap.php
```diff
@@ -28,6 +28,12 @@ class Imap
      */
     protected $tagCount = 0;
 
+    /**
+     * Last line returned from server
+     * @var string|null
+     */
+    protected $lastResponseLine;
```
glensc: use private unless needed higher visibility?

src/Protocol/Imap.php
```diff
@@ -143,9 +149,9 @@ protected function nextTaggedLine(&$tag)
         $line = $this->nextLine();
 
         // separate tag from line
-        list($tag, $line) = explode(' ', $line, 2);
+        list($tag, $this->lastResponseLine) = explode(' ', $line, 2);
```
glensc: use short syntax:
`[$tag, $this->lastResponseLine] = explode(' ', $line, 2);`

glensc (changes requested):

src/Protocol/Imap.php
```diff
@@ -821,4 +827,13 @@ public function search(array $params)
...
     }
+
+    /**
+     * @return string|null
+     */
+    public function getLastResponseLine()
+    {
+        return $this->lastResponseLine;
+    }
+
```
enumag:
```suggestion
```

PR #115, opened 2020-10-08
Fix str_split warning in ContentDisposition.php

smile1980:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

fix warning from str_split - there are cases when the value becomes less than 1, str_split tell "Warning: str_split(): The length of each segment must be greater than zero"


glensc:
@smile1980 include at least input email to reproduce the problem. and the pr title should describe what is being fixed.

weierophinney:
Obsolete following merge of #128

Issue #121, opened 2020-12-30
AddressList::addFromString method not fluent


braindawg:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.13.x

#### Summary

The docblock for the `AddressList::addFromString` method indicates that it should allow method chaining by returning an `AddressList`, but the method actually returns `void`.

#### Current behavior

`AddressList::addFromString` returns `void`, contrary to its docblock and the behavior of the rest of the state-modifying methods on `AddressList`. Therefore, attempting to chain any `AddressList` class methods after the `addFromString` call results in:

> Error: Call to a member function <chained_method> on null ...

#### How to reproduce

Attempt to chain a call to any `AddressList` method from the result of an `AddressList::addFromString` method, e.g.
`$addressList->addFromString('test@email.com')->has('test@email.com');`

Unit test will be added via a PR shortly.

#### Expected behavior

Per the docblock and the current behavior of all other state-modifying methods (`add`, `addMany`, `merge`), method chaining should work.

Ocramius:
Handled in #123

PR #122, opened 2020-12-30
Add unit test for #121: AddressList method fluency

braindawg:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR provides a failing unit test to demonstrate Issue #121.

### DIFF
--- a/test/AddressListTest.php
+++ b/test/AddressListTest.php
@@ -13,6 +13,7 @@ use Laminas\Mail\AddressList;
 use Laminas\Mail\Exception\InvalidArgumentException;
 use Laminas\Mail\Header;
 use PHPUnit\Framework\TestCase;
+use Throwable;
 
 /**
  * @group      Laminas_Mail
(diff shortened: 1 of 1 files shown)


test/AddressListTest.php
```diff
@@ -116,6 +117,18 @@ public function testCanAddManyAddressesAtOnce()
         $this->assertTrue($this->list->has('announce@example.com'));
     }
 
+    public function testCanAddFromStringFluently()
+    {
+        try {
+            $this->list->addFromString('test_fromstring_fluency1@example.com')
+                ->addFromString('test_fromstring_fluency2@example.com');
+        } catch (Throwable $e) {
```
Ocramius: This `catch` is not necessary

Ocramius (changes requested):
Let's target 2.12.x also here, since it's a bugfix

test/AddressListTest.php
braindawg: Took the cue on that from a couple other unit tests in AddressListTest.php that were similarly `catching` their errors. Removed mine!

Ocramius:
Thanks @braindawg!

PR #123, opened 2020-12-30
Fix issue #121: Correct return value

### LINKED ISSUE #121: AddressList::addFromString method not fluent

braindawg:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Current calls to AddressList::addFromString do not return a value, contrary to the docblock and behavior of other AddressList-modifying methods.

This returns the modified AddressList object to maintain a fluent
interface, as all the other state-modifying methods on AddressList do
(and as the docblock indicates that it should).

Fixes #121

### DIFF
--- a/src/AddressList.php
+++ b/src/AddressList.php
@@ -96,6 +96,7 @@ class AddressList implements Countable, Iterator
     public function addFromString($address, $comment = null)
     {
         $this->add(Address::fromString($address, $comment));
+        return $this;
     }
 
     /**


Ocramius (approved):
LGTM, but could you please target the 2.12.x branch?

glensc:
can https://github.com/laminas/laminas-mail/pull/108 be please taken care of as well?

PR #140, opened 2021-02-27
[2.14]: Fix: Fold long lines during SMTP communication

glensc:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

It seems, the underlying problem of the bug is that currently, the loop reads up to 1000 bytes, and if the input is longer than that, then everything over 1000 bytes will appear as the next line in mail headers, resulting in corruption. And depending on MTA, such corruption is treated as the end of headers input or adds another mail header.

Updates https://github.com/laminas/laminas-mail/pull/138 by @vladsf with the solution described in https://github.com/laminas/laminas-mail/pull/138#issuecomment-787063554 and adds unit test.

### DIFF
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -368,7 +368,7 @@ class Smtp extends AbstractProtocol
             if (strlen($line) > 998) {
                 // Long lines are "folded" by inserting "<CR><LF><SPACE>"
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
-                $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
+                $line = substr(chunk_split($line, 997, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
             }
 
             $this->_send($line);
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -8,6 +8,8 @@
 
 namespace Laminas\Mail\Protocol;
 
+use Laminas\Mail\Headers;
+
 /**
  * SMTP implementation of Laminas\Mail\Protocol\AbstractProtocol
  *
@@ -335,7 +337,7 @@ class Smtp extends AbstractProtocol
             if (strlen($line) > 998) {
                 // Long lines are "folded" by inserting "<CR><LF><SPACE>"
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
-                $line = substr(chunk_split($line, 998, "\r\n "), 0, -3);
+                $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
             }
             $this->_send($line);
         }
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -326,21 +326,42 @@ class Smtp extends AbstractProtocol
         unset($data);
         rewind($fp);
 
-        // max line length is 998 char + \r\n = 1000
-        // read the line up to PHP_SOCK_CHUNK_SIZE
-        while (($line = stream_get_line($fp, 0, "\n")) !== false) {
-            $line = rtrim($line, "\r");
+        $chunkSize = 4096;
+        $line = '';
+        while (($buffer = fgets($fp, $chunkSize)) !== false) {
+            $line .= $buffer;
+
+            // partial read, continue loop to read again to complete the line
+            if (strlen($buffer) === $chunkSize - 1 && $buffer[$chunkSize - 2] !== "\n") {
+                continue;
+            }
+
+            $line = rtrim($line, "\r\n");
             if (isset($line[0]) && $line[0] === '.') {
                 // Escape lines prefixed with a '.'
                 $line = '.' . $line;
             }
+
+            // max line length is 998 char + \r\n = 1000
             if (strlen($line) > 998) {
                 // Long lines are "folded" by inserting "<CR><LF><SPACE>"
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
                 $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
             }
             $this->_send($line);
+            $line = '';
         }
+        if ($line) {
+            // max line length is 998 char + \r\n = 1000
+            if (strlen($line) > 998) {
+                // Long lines are "folded" by inserting "<CR><LF><SPACE>"
+                // https://tools.ietf.org/html/rfc5322#section-2.2.3
+                $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
+            }
+
+            $this->_send($line);
+        }
+
         fclose($fp);
 
         $this->_send('.');
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -22,6 +22,12 @@ class Smtp extends AbstractProtocol
     use ProtocolTrait;
 
     /**
+     * RFC 5322 section-2.2.3 specifies maximum of 998 bytes per line.
+     * @see https://tools.ietf.org/html/rfc5322#section-2.2.3
+     */
+    public const MAX_LINE_LENGTH = 998;
+
+    /**
      * The transport method for the socket
      *
      * @var string
@@ -364,11 +370,13 @@ class Smtp extends AbstractProtocol
                 $line = '.' . $line;
             }
 
-            // max line length is 998 char + \r\n = 1000
-            if (strlen($line) > 998) {
+            if (strlen($line) > self::MAX_LINE_LENGTH) {
                 // Long lines are "folded" by inserting "<CR><LF><SPACE>"
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
-                $line = substr(chunk_split($line, 997, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
+                // Add "-1" to stay within limits,
+                // because Headers::FOLDING includes a byte for space character after \r\n
+                $chunks = chunk_split($line, self::MAX_LINE_LENGTH - 1, Headers::FOLDING);
+                $line = substr($chunks, 0, -strlen(Headers::FOLDING));
             }
 
             $this->_send($line);
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -331,7 +331,7 @@ class Smtp extends AbstractProtocol
                 $line .= $buffer;
 
                 // partial read, continue loop to read again to complete the line
-                if (strlen($buffer) === $chunkSize - 1 && $buffer[$chunkSize - 2] !== "\n") {
+                if (isset($buffer[$chunkSize - 2]) && $buffer[$chunkSize - 2] !== "\n") {
                     continue;
                 }
 
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -325,11 +325,17 @@ class Smtp extends AbstractProtocol
         rewind($fp);
 
         // max line length is 998 char + \r\n = 1000
-        while (($line = stream_get_line($fp, 1000, "\n")) !== false) {
+        // read the line up to PHP_SOCK_CHUNK_SIZE
+        while (($line = stream_get_line($fp, 0, "\n")) !== false) {
             $line = rtrim($line, "\r");
             if (isset($line[0]) && $line[0] === '.') {
                 // Escape lines prefixed with a '.'
                 $line = '.' . $line;
+            }
+            if (strlen($line) > 998) {
+                // Long lines are "folded" by inserting "<CR><LF><SPACE>"
+                // https://tools.ietf.org/html/rfc5322#section-2.2.3
+                $line = substr(chunk_split($line, 998, "\r\n "), 0, -3);
             }
             $this->_send($line);
         }
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -23,9 +23,10 @@ class Smtp extends AbstractProtocol
 
     /**
      * RFC 5322 section-2.2.3 specifies maximum of 998 bytes per line.
+     * This may not be exceeded.
      * @see https://tools.ietf.org/html/rfc5322#section-2.2.3
      */
-    public const MAX_LINE_LENGTH = 998;
+    public const SMTP_LINE_LIMIT = 998;
 
     /**
      * The transport method for the socket
@@ -387,12 +388,12 @@ class Smtp extends AbstractProtocol
                 $line = '.' . $line;
             }
 
-            if (strlen($line) > self::MAX_LINE_LENGTH) {
+            if (strlen($line) > self::SMTP_LINE_LIMIT) {
                 // Long lines are "folded" by inserting "<CR><LF><SPACE>"
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
                 // Add "-1" to stay within limits,
                 // because Headers::FOLDING includes a byte for space character after \r\n
-                $chunks = chunk_split($line, self::MAX_LINE_LENGTH - 1, Headers::FOLDING);
+                $chunks = chunk_split($line, self::SMTP_LINE_LIMIT - 1, Headers::FOLDING);
                 $line = substr($chunks, 0, -strlen(Headers::FOLDING));
             }
 
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -317,25 +317,36 @@ class Smtp extends AbstractProtocol
         $this->_send('DATA');
         $this->_expect(354, 120); // Timeout set for 2 minutes as per RFC 2821 4.5.3.2
 
-        if (($fp = fopen("php://temp", "r+")) === false) {
-            throw new Exception\RuntimeException('cannot fopen');
-        }
-        if (fwrite($fp, $data) === false) {
-            throw new Exception\RuntimeException('cannot fwrite');
-        }
-        unset($data);
-        rewind($fp);
+        $chunkReader = static function (string $data, int $chunkSize = 4096) {
+            if (($fp = fopen("php://temp", "r+")) === false) {
+                throw new Exception\RuntimeException('cannot fopen');
+            }
+            if (fwrite($fp, $data) === false) {
+                throw new Exception\RuntimeException('cannot fwrite');
+            }
+            rewind($fp);
 
-        $chunkSize = 4096;
-        $line = '';
-        while (($buffer = fgets($fp, $chunkSize)) !== false) {
-            $line .= $buffer;
+            $line = null;
+            while (($buffer = fgets($fp, $chunkSize)) !== false) {
+                $line .= $buffer;
 
-            // partial read, continue loop to read again to complete the line
-            if (strlen($buffer) === $chunkSize - 1 && $buffer[$chunkSize - 2] !== "\n") {
-                continue;
+                // partial read, continue loop to read again to complete the line
+                if (strlen($buffer) === $chunkSize - 1 && $buffer[$chunkSize - 2] !== "\n") {
+                    continue;
+                }
+
+                yield $line;
+                $line = null;
             }
 
+            if ($line !== null) {
+                yield $line;
+            }
+
+            fclose($fp);
+        };
+
+        foreach ($chunkReader($data) as $line) {
             $line = rtrim($line, "\r\n");
             if (isset($line[0]) && $line[0] === '.') {
                 // Escape lines prefixed with a '.'
@@ -348,21 +359,9 @@ class Smtp extends AbstractProtocol
                 // https://tools.ietf.org/html/rfc5322#section-2.2.3
                 $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
             }
-            $this->_send($line);
-            $line = '';
-        }
-        if ($line) {
-            // max line length is 998 char + \r\n = 1000
-            if (strlen($line) > 998) {
-                // Long lines are "folded" by inserting "<CR><LF><SPACE>"
-                // https://tools.ietf.org/html/rfc5322#section-2.2.3
-                $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
-            }
 
             $this->_send($line);
         }
-
-        fclose($fp);
 
         $this->_send('.');
         $this->_expect(250, 600); // Timeout set for 10 minutes as per RFC 2821 4.5.3.2
--- a/src/Protocol/Smtp.php
+++ b/src/Protocol/Smtp.php
@@ -8,6 +8,7 @@
 
 namespace Laminas\Mail\Protocol;
 
+use Generator;
 use Laminas\Mail\Headers;
 
 /**
@@ -173,6 +174,44 @@ class Smtp extends AbstractProtocol
     }
 
     /**
+     * Read $data as lines terminated by "\n"
+     *
+     * @param string $data
+     * @param int $chunkSize
+     * @return Generator|string[]
+     * @author Elan Ruusamäe <glen@pld-linux.org>
+     */
+    private static function chunkedReader(string $data, int $chunkSize = 4096): Generator
+    {
+        if (($fp = fopen("php://temp", "r+")) === false) {
+            throw new Exception\RuntimeException('cannot fopen');
+        }
+        if (fwrite($fp, $data) === false) {
+            throw new Exception\RuntimeException('cannot fwrite');
+        }
+        rewind($fp);
+
+        $line = null;
+        while (($buffer = fgets($fp, $chunkSize)) !== false) {
+            $line .= $buffer;
+
+            // partial read, continue loop to read again to complete the line
+            if (isset($buffer[$chunkSize - 2]) && $buffer[$chunkSize - 2] !== "\n") {
+                continue;
+            }
+
+            yield $line;
+            $line = null;
+        }
+
+        if ($line !== null) {
+            yield $line;
+        }
+
+        fclose($fp);
+    }
+
+    /**
      * Whether or not send QUIT command
      *
      * @return bool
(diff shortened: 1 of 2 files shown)


src/Protocol/Smtp.php
```diff
@@ -324,15 +326,35 @@ public function data($data)
         unset($data);
         rewind($fp);
 
-        // max line length is 998 char + \r\n = 1000
-        while (($line = stream_get_line($fp, 1000, "\n")) !== false) {
-            $line = rtrim($line, "\r");
+        $chunkSize = 4906;
```
vladsf: you mean 4096?

src/Protocol/Smtp.php
```diff
@@ -324,15 +326,35 @@ public function data($data)
...
+            // max line length is 998 char + \r\n = 1000
+            if (strlen($line) > 998) {
+                // Long lines are "folded" by inserting "<CR><LF><SPACE>"
+                // https://tools.ietf.org/html/rfc5322#section-2.2.3
+                $line = substr(chunk_split($line, 998, Headers::FOLDING), 0, -strlen(Headers::FOLDING));
+            }
             $this->_send($line);
+            $line = '';
         }
+        if ($line) {
```
vladsf: I guess this line requires escaping, rtrim'ing and folding too?

test/Transport/SmtpTest.php
```diff
@@ -199,6 +199,33 @@ public function testReceivesMailArtifacts(): void
         $this->assertStringContainsString("\r\n\r\nThis is only a test.", $data, $data);
     }
 
+    /**
+     * Fold long lines during smtp communication in Protocol\Smtp class.
+     * Test folding of long lines following RFC 5322 section-2.2.3
+     *
+     * @see https://github.com/laminas/laminas-mail/pull/138
```
vladsf: pull/140 to be precise

src/Protocol/Smtp.php
vladsf: I guess do { } while(fgets..); fits better in this case.

src/Protocol/Smtp.php
```diff
@@ -324,15 +326,35 @@ public function data($data)
...
-        // max line length is 998 char + \r\n = 1000
-        while (($line = stream_get_line($fp, 1000, "\n")) !== false) {
-            $line = rtrim($line, "\r");
+        $chunkSize = 4906;
+        $line = '';
+        while (($buffer = fgets($fp, $chunkSize)) !== false) {
+            $line .= $buffer;
+
+            // partial read, continue loop to read again to complete the line
+            if (strlen($buffer) === $chunkSize - 1 && $buffer[$chunkSize - 2] !== "\n") {
```
vladsf: I don't understand this. One has read full chunk 4095, no line break found. Why to look for "\n" at 4094?

glensc: That is for the scenario when the line is exactly of 4094 length, i.e not truncated.

For example, these both end up read of 4 bytes:
1. "123\n"
2. "1234"

but only the second example is "truncated", first one is exactly of the length of the buffer-1.

I tried to make testcase for this, but ended up finding another bug on Storage\Message not loading the wrapped headers back properly (injects ' ' at one (but not all!!!) wrap point).

test/Transport/SmtpTest.php
glensc: Yeah, I did not plan to make new PR when I wrote the unit test :)

src/Protocol/Smtp.php
glensc: Indeed, wrapping needed as well. Good catch! Yes, the code became quickly nasty.

I would add a new class dealing with reading and wrapping, but not sure if such refactoring would be accepted then. Some Generator that the original code can consume.

src/Protocol/Smtp.php
glensc: Absolutely correct! Nice catch!

src/Protocol/Smtp.php
vladsf: fgets does not truncate "\n":
```
$ for i in {1..4094}; do echo -n '1' >>inputfile.txt; done
$ php fgets.php 
4094 49
$ >inputfile.txt
$ for i in {1..4093}; do echo -n '1' >>inputfile.txt; done
$ echo '' >>inputfile.txt
$ php fgets.php 
4094 10
```

```
<?php
$handle = @fopen("inputfile.txt", "r");
if ($handle) {
	while (($buffer = fgets($handle, 4096)) !== false) {
		echo strlen($buffer);
		echo " " . ord($buffer[strlen($buffer)-1]);
		echo "\n";
	}
```
so I'm **not** sure `$buffer[$chunkSize - 2] !== "\n" `is correct.

glensc: @vladsf, will check your input later, I wrote code based on `fgets` documentation. Did not test properly. Could be simple off by one issue here. 

I got distracted with `Storage\Message` bug, that result wasn't decoded properly.

glensc: @vladsf indeed, the logic is inverted. if the line is incomplete the last char is not `\n`, otherwise, it is `\n`:

```php
#!/usr/bin/env php
<?php
$fname = "inputfile.txt";
file_put_contents($fname, str_repeat("A", 4096). "\n");

$handle = fopen($fname, "r") or die;

$chunkSize = 1024;
while (($buffer = fgets($handle, $chunkSize)) !== false) {
    $length = strlen($buffer);
    $c = $buffer[$length - 1];
    printf("len:%d of %d c=%s (ord:0x%x)\n", $length, $chunkSize, json_encode($c), ord($c));
}
```

```
len:1023 of 1024 c="A" (ord:0x41)
len:1023 of 1024 c="A" (ord:0x41)
len:1023 of 1024 c="A" (ord:0x41)
len:1023 of 1024 c="A" (ord:0x41)
len:5 of 1024 c="\n" (ord:0xa)
```

glensc: > so I'm not sure $buffer[$chunkSize - 2] !== "\n" is correct.

it is correct. `fgets` reads maximum of `$chunkSize` bytes, if line is complete last byte is "\n". accessing last byte of `$buffer`, is at `strlen($buffer)-1` offset, so, if maximum read happens, then also `strlen($buffer) == $chunkSize-1`, so `strlen($buffer)-1` can be replaced with `$chunkSize-2`.

The condition could be replaced with:
```php
if (isset($buffer[$chunkSize - 2]) && $buffer[$chunkSize - 2] !== "\n") {
```

but not sure, maybe that's harder to read and understand?

src/Protocol/Smtp.php
glensc: having `do {} while (fgets)`  gets awkward, as then need to check in the loop if $buffer is even initialized. so I'll use Generator, and do one yield after the loop than make the loop itself complicated.

See e80773f6 commit. I ended up reset `$line` to `null` to keep the support of sending empty lines (length=0).

glensc:
I would move `$chunkReader function` to a new class (preferred) or to this class method. But really need some guidance from The Maintainers.

src/Protocol/Smtp.php
vladsf: It seems to me that is hard to read and triggers "array index out of bounds error" in my mind.
Could it be like this? -

```
if (strlen($buffer) == $chunkSize-1) {
       if ($buffer[strlen($buffer) - 1] !== "\n") {
..
```

vladsf:
I honestly believe that too much PHP force has been used :)

vladsf:
> I would move `$chunkReader function` to a new class (preferred) or to this class method. But really need some guidance from The Maintainers.

That was an impressive technique :)  Wouldn't be simpler to move "escaping, rtrim'ing and folding" a line into a function?

src/Protocol/Smtp.php
glensc: the out of bounds access is protected with `isset`, and as the calculation is known ahead, there's optimization to avoid costly `strlen()` call in the loop.

there's a similar `isset($foo[0])` in the code, so it's not that mystical.

and in answer to you question, yes. the logic is the same. extra `if` with `strlen($buffer) == $chunkSize-1` is identical in behavior with `isset($buffer[$chunkSize - 2]) &&`

glensc:
> That was an impressive technique :) Wouldn't be simpler to move "escaping, rtrim'ing and folding" a line into a function?

what do you mean by "simpler", simpler to whom? to what? i offered to move the generator to class method, but not doing 'jus for that', as maybe maintainers accept it as is now.

also, I would be against putting three different logics to the same code block, it would make code more difficult to understand and follow the code flow. currently, it's distinct, reading and unfolding in one place, the rest in the other loop. the code gets smelly if they all done in the same loop (as you can see from commit history). need multiple state variables that are unrelated to each other but used in the same loop.

src/Protocol/Smtp.php
vladsf: Yes, the logic is the same. I'm only sharing my readability concerns.

vladsf:
> what do you mean by "simpler", simpler to whom? to what? i offered to move the generator to class method, but not doing 'jus for that', as maybe maintainers accept it as is now.

For the maintainers to review the PR.

glensc:
> > what do you mean by "simpler", simpler to whom? to what? I offered to move the generator to class method, but not doing 'jus for that', as maybe maintainers accept it as is now.
> 
> For the maintainers to review the PR.

In the end, the resulting code is simpler, and only the aggregated diff is messy, but individual commit diffs are okay.

glensc:
Created reproducer of the bug I discovered that prevented of making more reasonable tests:
- https://github.com/laminas/laminas-mail/pull/142


Wanted to test like this:

```php
        $raw = substr($data, strpos($data, "DATA\r\n") + 6);

        $message = new StorageMessage(['raw' => $raw]);
        $headers = $message->getHeaders()->toArray();
        $this->assertEquals($bufferSizeHeaderValue, $headers['X-Exact-Length']);
        $this->assertEquals($headerValue, $headers['X-Ms-Exchange-Antispam-Messagedata']);
```

That's what I get when captured SMTP from this PR. It does look correct, so it's a bug in parsing of this library/

glensc:
Thought this could go with 2.13, but then saw that adapt to new CI flow would be missing, so changed back to 2.14.

glensc:
@Ocramius can you do the review, merge, and release dance? thank you!

test/Transport/SmtpTest.php
```diff
@@ -199,6 +199,40 @@ public function testReceivesMailArtifacts(): void
...
+        $message->getHeaders()->addHeaders([
+            'X-Ms-Exchange-Antispam-Messagedata' => $headerValue,
+            'X-Exact-Length' => $bufferSizeHeaderValue,
+        ]);
+
+        $this->transport->send($message);
+        $data = $this->connection->getLog();
+
+        $lines = explode("\r\n", $data);
+        foreach ($lines as $line) {
```
Slamdunk, line 231: In the way the test is written, somehow someday `$data` could turn empty, and the assertions inside `foreach` go silently ignored. Please add an assertion to verify there's a content, the more precise the better:

```suggestion
        $this->assertCount(123, $lines);
        foreach ($lines as $line) {
```

Slamdunk (changes requested):

glensc:
@vladsf I've changed the chunk_split argument to 997, as leading space on the "new" line, also adds a byte:

- Input: `"1234567890123456\n"`
- Split: `"123456789\r\n 0123456\n"`
  - Line1; `"123456789\r\n`
  - Line2: `" 0123456\n"` (note the leading space)

This came up after I fixed the test not to accept anything over 998.

test/Transport/SmtpTest.php
glensc: @Slamdunk good catch! Committed.

test/Transport/SmtpTest.php
```diff
@@ -199,6 +199,43 @@ public function testReceivesMailArtifacts(): void
...
+     */
+    public function testLongLinesFoldingRFC5322(): void
+    {
+        $message = $this->getMessage();
+        // Create buffer of 8192 bytes (PHP_SOCK_CHUNK_SIZE)
+        $buffer = str_repeat('0123456789abcdef', 512);
+        $this->assertEquals(8192, strlen($buffer));
+
+        $headerValue = substr($buffer, 0, 1000);
+        $bufferSizeHeaderValue = substr($buffer, 0, 998 - strlen('X-Exact-Length') - 2);
```
Slamdunk: The code looks good, but I'm puzzled by all these hardcoded numbers: 1000, 998, 997, 2, both in the tests and in the src.
Your comments to the code are greats, but I fear that if someone will ever change one of these, another one get forgotten.

Could you try to refactor all of them into verbose `Protocol\Smtp` constants and combinations of them?

Slamdunk (changes requested):

test/Transport/SmtpTest.php
glensc: Well, that's for the unit tests exist, right?

In overall:
- the line length must be at most 1000 bytes
- the line must be terminated with `\r\n`, so it's 998
- now the 997 for chunker, comes from the fact that `Headers::FOLDING` inserts space at the new line, see this comment: https://github.com/laminas/laminas-mail/pull/140#issuecomment-791226698

I can add constant 1000, and use -2 and -1, but these rely on what exactly is contained in Headers::FOLDING constant. so if someone changes that constant the calculations would be off again.

and making the logic super smart by inspecting what is in Headers::FOLDING seems counterproductive as the constant will never be changed, and if it's changed there are tons of places that can go wrong.

so I'm, contradicted what constants to add, as adding 3 constants seems odd as well.

glensc: Added MAX_LINE_LENGTH constant. Please mark discussion resolved if this is an acceptable outcome.

Slamdunk: I would be perfectly fine having `strlen(Headers::FOLDING)` everywhere, because this code is tricky, and readability should come first.
I would also like a `const SPACE_LENGTH = 1`: the more readable the code is, the easier it is to understand

vladsf: @glensc @Slamdunk Please consider renaming MAX_LINE_LENGTH to SMTP_LINE_LIMIT. This is not a library's but SMTP protocol limit.

vladsf: @glensc Changing to 997 is correct! Good fix!

Slamdunk: > Please consider renaming MAX_LINE_LENGTH to SMTP_LINE_LIMIT. This is not a library's but SMTP protocol limit.

:+1:

glensc: @Slamdunk `strlen(Headers::FOLDING)` is incorrect and source of new bugs. it depends that the folding has one space after `"\r\n"`. `SPACE_LENGTH` seems also a pointless constant that's been used only once.

also, the strlen is not needed, as the RFC explicitly says 998 must be used.

I rather add a test that tests that `Headers::FOLDING` has exactly the current value.

glensc: > @glensc @Slamdunk Please consider renaming MAX_LINE_LENGTH to SMTP_LINE_LIMIT. This is not a library's but SMTP protocol limit.

Not convinced that's necessary, the constant is in class Protocol\SMTP, so it's already in code specific to SMTP protocol... and it's accompanied with a comment and with a link to specification.

Slamdunk: > @Slamdunk `strlen(Headers::FOLDING)` is incorrect and source of new bugs. it depends that the folding has one space after `"\r\n"`. `SPACE_LENGTH` seems also a pointless constant that's been used only once.
> 
> also, the strlen is not needed, as the RFC explicitly says 998 must be used.

Ok



> I rather add a test that tests that `Headers::FOLDING` has exactly the current value.

That would help



> Not convinced that's necessary, the constant is in class Protocol\SMTP, so it's already in code specific to SMTP protocol... and it's accompanied with a comment and with a link to specification.

Ok


Still it's hard to me to understand what the `2` here stands for:

https://github.com/laminas/laminas-mail/pull/140/files#diff-1aa66d5ccd570f8fab05c2a05fe3b18ae45d4866faf5e83dcd64993a04d0ba2aR205

Why `2` and not `1`? Please reply not here but in the code

vladsf: > Not convinced that's necessary, the constant is in class Protocol\SMTP, so it's already in code specific to SMTP protocol... and it's accompanied with a comment and with a link to specification.

Well, I am missing the purpose of code review then if there is only one way to go. People read code many times, why not to help them to avoid reading comments?? SMTP_LINE_LIMIT clearly states it can not be increased while MAX_LINE_LENGTH might be. 

More on this topic, I do _not_  think the patch needs to be so complex. RFC only states the line must be 998 characters long at max, but it does not state _where_ longs lines have to broken. E.g. I don't think the library has to read an endless line until "\n" is met or PHP runs out of memory. There is no much reason to read a 1MB line and then split it. _Default_ 8K is good enough.

glensc: > Well, I am missing the purpose of code review then if there is only one way to go. 

I disagreed with the constant need to include SMTP in it, with a counterargument, the other reviewer accepted that. The code review can go endlessly if every now someone throws something in, conflicting with previous review notes.

> People read code many times, why not to help them to avoid reading comments?? SMTP_LINE_LIMIT clearly states it can not be increased while MAX_LINE_LENGTH might be.

apparently, you did not explain yourself clearly initially, seemed to me you argued about protocol SMTP not being in the constant name, but now seems you mean about MAX vs LIMIT. now I see the difference.

Also, Why do I write code comments if nobody reads them?

> More on this topic, I do not think the patch needs to be so complex. 

In my opinion, the current code with the generator is simplex rather than complex, the reader deals with reading in chunks emitting full lines. reader consumer always gets a full line.

> RFC only states the line must be 998 characters long at max, but it does not state where longs lines have to broken. E.g. I don't think the library has to read an endless line until "\n" is met or PHP runs out of memory. There is no much reason to read a 1MB line and then split it. Default 8K is good enough.

I strongly disagree with statements like "Some value is good enough", as that's what got this into first place, someone thought earlier 1000 is good enough, but that resulted in longer lines than 1000 a data corruption, so your idea just to up limit will do exactly that, for someone sending 9K line will cause a corruption:
- https://github.com/laminas/laminas-mail/pull/138#issuecomment-787063554

your solution did not deal with the problem that if a line is partially read (longer than buffer size), a proper folding is added.

glensc: > Why 2 and not 1? Please reply not here but in the code

it's combining -1 to access last char, and -1 because how how `fgets()` work:
> `fgets()`: Returns a string of up to `length-1` bytes read from the file pointed to by handle. If there is no more data to read in the file pointer, then false is returned.

- https://www.php.net/fgets


see the previous discussion:
- https://github.com/laminas/laminas-mail/pull/140#discussion_r584196792

can you help, how to explain it in the code?

adding intermediate variables could help, but it's not good practice to do such things in a loop just for readability...

do you feel better if the code calls `fgets(...$buffsize+1)`, so the check would be `$bufsiz-1`?

glensc: here's something i've came up with:

```php
while (($buffer = fgets($handle, $chunkSize)) !== false) {


    // This is optimization to avoid calling length() in a loop.
    // We need to match a condition that is when:
    // 1. maximum was read from fgets, which is $chunkSize-1
    // 2. last byte of the buffer is not \n
    //
    // when maximum is read, the last byte of buffer is at $chunkSize-2:
    // - strlen($buffer) === $chunkSize-1
    // - strlen($buffer)-1 === $chunkSize-2
    // - $buffer[strlen($buffer) - 1] === $buffer[$chunkSize-2]
    if (isset($buffer[$chunkSize - 2]) && $buffer[$chunkSize - 2] !== "\n") {
        printf("x1: -> line is truncated\n");
    }
```

cold also do this, but this is against doing "useless" things in a loop:

```php
    $lastByte = $buffer[$chunkSize - 2] ?? null;
    if ($lastByte !== "\n") {
        printf("x1: -> line is truncated\n");
    }
```

vladsf: > I strongly disagree with statements like "Some value is good enough", as that's what got this into first place, someone thought earlier 1000 is good enough, but that resulted in longer lines than 1000 a data corruption, so your idea just to up limit will do exactly that, for someone sending 9K line will cause a corruption:
> 
>     * [#138 (comment)](https://github.com/laminas/laminas-mail/pull/138#issuecomment-787063554)
> 
> 
> your solution did not deal with the problem that if a line is partially read (longer than buffer size), a proper folding is added.

Closed PR #138 proposes a correct folding. That comment mentioned was a temporary workaround.

I guess PHP_SOCK_CHUNK_SIZE was set 8K for a purpose (unknown to me though). May be for for quicker read-send-read-send operations. SMTP communication does not requires all broken lines to of equal size. It can accept and "unfold" lines of any size later. This example folded lines are equal once "unfolded":
```
<SPACE>ABCDEF
<SPACE>ABCDEF
<SPACE>ABCDEF
```
```
<SPACE>AB
<SPACE>CDEF
<SPACE>ABCD
<SPACE>EFABCDEF
```
So I am trying to explain that sending folded 8K string is good enough and there is little win in reading e.g. 800K line and then fold'ing it only one has been fully read.

Slamdunk (approved):

test/Transport/SmtpTest.php
vladsf: the lines I suggested looks more readable to me. One checks the buffer size and tests the last byte. 
```
if (strlen($buffer) == $chunkSize-1) { 
       if ($buffer[strlen($buffer) - 1] !== "\n") {
..
```

vladsf: readability is more important than strlen/isset speed difference.

glensc: > So I am trying to explain that sending folded 8K string is good enough and there is little win in reading e.g. 800K line and then fold'ing it only one has been fully read.

and I'm saying "good enough" is not a fix, but a workaround. your code does not distinguish between a truncated and complete line.

glensc: > the lines I suggested looks more readable to me. One checks the buffer size and tests the last byte.
> readability is more important than strlen/isset speed difference.

There's no need to repeat same text over and over! I replied then and I reply now, not going o change if maintainers do not request such change, and so far they have not given their opinion on the matter. and to you it may be readable, to me, it's even harder to read.

vladsf: > and I'm saying "good enough" is not a fix, but a workaround. your code does not distinguish between a truncated and complete line.

There is no reason to read full line which exceeds 8K chracters. It only adds a vulnerability.

vladsf: > can you help, how to explain it in the code?

You asked so I answered.

glensc: >> can you help, how to explain it in the code?
>
> You asked so I answered.

No, [that][1] was reply to @Slamdunk [question][2]

[2]: https://github.com/laminas/laminas-mail/pull/140#discussion_r588191660
[1]: https://github.com/laminas/laminas-mail/pull/140#discussion_r588243003

src/Protocol/Smtp.php
```diff
@@ -170,6 +180,61 @@ public function setUseCompleteQuit($useCompleteQuit)
...
+            // to access last byte of buffer, we can do
+            // - $buffer[strlen($buffer)-1]
+            // and when maximum is read from fgets, then:
+            // - strlen($buffer) === $chunkSize-1
+            // - strlen($buffer)-1 === $chunkSize-2
+            // which means this is also true:
+            // - $buffer[strlen($buffer)-1] === $buffer[$chunkSize-2]
+            //
+            // the null coalesce works, as string offset can never be null
+            $lastByte = $buffer[$chunkSize - 2] ?? null;
```
vladsf, line 219: Sorry, lastByte means the last byte and it can't be Null. The name of the variable contradicts its value.

src/Protocol/Smtp.php
```diff
@@ -170,6 +180,61 @@ public function setUseCompleteQuit($useCompleteQuit)
...
+
+            // partial read, continue loop to read again to complete the line
+            // compare \n first as that's usually false
+            if ($lastByte !== "\n" && $lastByte !== null) {
+                continue;
+            }
```
vladsf, line 225: Why to reinvent reading a line if `fgets($fp)` will keep reading from the stream until it reaches the end of the line?
Just as simple as -
```
while (($buffer = fgets($fp)) !== false) {
  yield $buffer;
}
```

glensc: @vladsf I've explained this to you multiple times, and I'm tired already, and for the last time: that's the underlying problem, such code can't distinguish between whole line and partial line, resulting in corruption that this PR is fixing.

run the code if you don't understand the explanations and don't understand how fgets works and why the switch was made to fgets (I've explained it!).

src/Protocol/Smtp.php
glensc: I don't see you suggesting a better name. What's the point of such reviews?

> "There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors." – Leon Bambrick

- https://www.mediawiki.org/wiki/Naming_things

src/Protocol/Smtp.php
vladsf: Surely I forgot this _read-the-full-line_ nuance, you are not the one feeling tired. You removed my comment from initial commit which had this explained this `// read the line up to PHP_SOCK_CHUNK_SIZE` and it's only a week has passed. Imagine another contributor looking into the code and having no clue why reading using $chunkSize and why the $lastByte is always null if $chunkSize > PHP_SOCK_CHUNK_SIZE.

src/Protocol/Smtp.php
vladsf: I need a break for ideas.

src/Protocol/Smtp.php
glensc: `// read the line up to PHP_SOCK_CHUNK_SIZE` comment is removed because it's bogus to keep, the current code does behave that way. I've explained it in the notes in the PRs before making the change, and after I've made the change. I'm tired of explaining the same thing over and over to you, which you just discard.

glensc: > [*snip*] if $chunkSize > PHP_SOCK_CHUNK_SIZE.

`PHP_SOCK_CHUNK_SIZE` is irrelevant in any way, comment on the current code, not the past code. see the commits, see the diffs.

src/Protocol/Smtp.php
vladsf: @glensc, @Slamdunk  I suggest following readability improvement which speaks for itself without comments
(it replaces lines 204-227):
```
$lineBreak = $buffer[-1] ?? null;
if ($lineBreak === "\n") {
  yield $line;
  $line = null;
}
```

src/Protocol/Smtp.php
```diff
@@ -170,6 +179,61 @@ public function setUseCompleteQuit($useCompleteQuit)
...
 
+    /**
+     * Read $data as lines terminated by "\n"
+     *
+     * @param string $data
+     * @param int $chunkSize
+     * @return Generator|string[]
+     * @author Elan Ruusamäe <glen@pld-linux.org>
+     */
+    private static function chunkedReader(string $data, int $chunkSize = 4096): Generator
```
vladsf, line 191: Test method is named `testLongLinesFoldingRFC5322`. @glensc @Slamdunk I suggest renaming `chunkedReader` to `longLineReader` so it conforms with the test method name and its primary purpose - to read long lines complete. It does not return equal size chunks, it returns lines of random length.

glensc:
@Slamdunk made a fresh rebase against `upstream/2.14.x`, perhaps it's better to review knowing it's up to date with the base branch.

Slamdunk:
Thank you @glensc, I'm going to squash this PR to reduce `git-blame` and `git-bisect` verbosity

glensc:
@Slamdunk perhaps you should have asked first from authors for permission to squash!

This PR has waited for two weeks+ being idle! If this was an issue, should have said it in that time period!

I (try to) follow the atomic commits principle:
- https://www.freshconsulting.com/atomic-commits/

As I see it, you made the opposite: more difficult to git blame/bisect with squashing. And losing commit author information is not very friendly either.

glensc:
Update: seems this got merged without squashing, at least the 542686ae merge commit. So far so good.

glensc:
@Slamdunk thank you for the 2.14.0 release!

glensc:
Seems the release automation can't handle that PR had multiple contributors:
- https://github.com/laminas/laminas-mail/releases/tag/2.14.0

> 140: [2.14]: Fix: Fold long lines during SMTP communication thanks to @glensc

it actually replaced #138 by @vladsf yet there's no credit about that.

PR #153, opened 2021-05-19
Fix #152: Use strict string comparisons for header keys

### LINKED ISSUE #152: Headers#has() can return true when it should return false
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | HEAD (2.15.x)

#### Summary

`Laminas\Mail\Headers#has()` will return `true` for some values when it should return `false`

#### Current behavior

If a `Headers` instance contains a header with key `'00'`, `$headers->has('00000')` will return `true`.

#### How to reproduce

```php
$headers = new Headers();
$headers->addHeaderLine('0', 'test');
$headers->addHeaderLine('1', 'test');

var_export($headers->has('00000'));  // true
echo PHP_EOL;
var_export($headers->has('00001'));  // true
```

#### Expected behavior

```php
$headers = new Headers();
$headers->addHeaderLine('0', 'test');
$headers->addHeaderLine('1', 'test');

var_export($headers->has('00000'));  // false
echo PHP_EOL;
var_export($headers->has('00001'));  // false
```

Zenexer:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fixes #152 by using strict string comparisons with `in_array` and `array_keys`.

### DIFF
--- a/src/Address.php
+++ b/src/Address.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-mail for the canonical source repository
- * @copyright https://github.com/laminas/laminas-mail/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-mail/blob/master/LICENSE.md New BSD License
- */
-
 namespace Laminas\Mail;
 
 use Laminas\Validator\EmailAddress as EmailAddressValidator;
--- a/src/Address/AddressInterface.php
+++ b/src/Address/AddressInterface.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-mail for the canonical source repository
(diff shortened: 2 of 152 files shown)


glensc:
Some foreign commit in the pr 0b4742f6a624a970c300850d09cac618d05ae405, if it's not an error, should be clearly indicated in PR body why it's here and the state changed to Draft.

Zenexer:
@glensc, that's because @Xerkus changed the target branch, so it's carrying over commits from 2.15.x, if I'm not mistaken.  My original PR only had one commit.  I can cherry-pick it onto 2.14.x if that's preferable.

Xerkus:
Yes, you would need to rebase onto 2.14.x since this is a bugfix and next patch release is 2.14.1

Zenexer:
Replacing this with #154.

PR #163, opened 2021-08-18
Fix encoding in sendmail transport with php8

### LINKED ISSUE #162: [BUG] [PHP8] "Sendmail" Transport - Mailencoding is broken after php core fixes
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.14.X + 2.15.X
| PHP |  >= 8.0

#### Summary

PHP Core fixed a bug in php ( https://bugs.php.net/bug.php?id=47983 ) which expectes mail content to contain CRLF line endings 

#### Current behavior

https://github.com/laminas/laminas-mail/blob/ee318f54ca9d9d475acfea074e8077e56c5198bc/src/Transport/Sendmail.php#L123-L128

Line endings are filtered

more infos on topic: https://github.com/PHPMailer/PHPMailer/pull/2188

#### How to reproduce

Send an HTML email with `sendmail` transport with PHP 7 and PHP 8 and see the PHP 8 versions is broken because it can't be parsed and is detected as `Content-Type: text/plain`

#### Expected behavior

https://github.com/laminas/laminas-mail/blob/ee318f54ca9d9d475acfea074e8077e56c5198bc/src/Transport/Sendmail.php#L123-L128

don't replace `"\r\n"` with `"\n"`

#### Fix:

remove the replaces (PR is #164  )

Fahl-Design:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

fixes #162

### DIFF
--- a/test/Storage/MboxFolderTest.php
+++ b/test/Storage/MboxFolderTest.php
@@ -304,6 +304,7 @@ class MboxFolderTest extends TestCase
         } catch (\Exception $e) {
             $check = true;
             // test ok
+            $this->assertTrue($check);
         }
 
         chmod($this->params['dirname'] . 'subfolder', $stat['mode']);
--- a/test/Storage/MboxFolderTest.php
+++ b/test/Storage/MboxFolderTest.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-mail for the canonical source repository
- * @copyright https://github.com/laminas/laminas-mail/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-mail/blob/master/LICENSE.md New BSD License
- */
-
 namespace LaminasTest\Mail\Storage;
 
 use Laminas\Config;
--- a/test/Storage/MboxTest.php
+++ b/test/Storage/MboxTest.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-mail for the canonical source repository
- * @copyright https://github.com/laminas/laminas-mail/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-mail/blob/master/LICENSE.md New BSD License
- */
-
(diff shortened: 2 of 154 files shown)


test/Storage/MboxFolderTest.php
```diff
@@ -304,6 +304,7 @@ public function testNotReadableFolder(): void
         } catch (\Exception $e) {
             $check = true;
             // test ok
+            $this->assertTrue($check);
```
Slamdunk, line 307: This should be outside the `try-catch`

test/Storage/MboxTest.php
```diff
@@ -276,6 +276,7 @@ public function testSleepWakeRemoved(): void
         } catch (\Exception $e) {
             $check = true;
             // test ok
+            $this->assertTrue($check);
```
Slamdunk, line 279: This too

Slamdunk (changes requested):
Awesome, thank you. Just a little fix and we are done

Slamdunk:
Also, please rebase it onto `2.14.x` branch as this is a bugfix

test/Storage/MboxTest.php
Fahl-Design: Since psalm is getting nervous when asserting `assertTrue` after the try-catch block, i changed to `assertFileExist` is useless anyway just to prevent phpunit "risky test" ;)

test/Storage/MboxFolderTest.php
Fahl-Design: Since psalm is getting nervous when asserting `assertTrue` after the try-catch block, i changed to `assertFolderExist`  is useless anyway just to prevent phpunit "risky test" ;)

Fahl-Design:
hi @Slamdunk thanks for reviewing, I created a new PR #164 based on `2.14.x`, this can be closed

PR #166, opened 2021-09-20
Provide PHP 8.1 support

weierophinney:
This patch provides PHP 8.1 support for laminas-mail via the following changes:

- Adds `~8.1.0` to the list of allowed PHP versions
- Bumps dependencies to versions known to work with 8.0/8.1 where possible
- Removes laminas-config as a dependency
  - It was a soft dependency previously; the source code did not directly reference it.
  - Refactoring of storage adapters was done to normalize options/params to associative arrays; this will also work on laminas-config instances, providing backwards compatibility.
- Changes how the package replaces zend-mail
  - Renames "replace" section to "conflict"
  - Changes "zend-mail" constraint in that section to "*"
  - Removes dependency on laminas-zendframework-bridge
- Adds `#[ReturnTypeWillChange]` attributes where needed
- When ext-intl is present, we now use `idn_to_ascii()` for punycode conversion, as the true/punycode package raises errors due to "implicit float to int conversion" under 8.1, leading to broken punycode conversion.
- `list() = explode()` raises a warning under 8.1 if an index is not created, vs a notice in previous versions; the patch ~~now detects this~~ refactors these to test for delimiters before attempting to explode a string.

### DIFF
--- a/src/Storage/Folder/Mbox.php
+++ b/src/Storage/Folder/Mbox.php
@@ -131,10 +131,15 @@ class Mbox extends Storage\Mbox implements FolderInterface
         $currentFolder = $this->rootFolder;
         $subname = trim($rootFolder, DIRECTORY_SEPARATOR);
         while ($currentFolder) {
-            ErrorHandler::start(E_NOTICE | E_WARNING);
-            list($entry, $subname) = explode(DIRECTORY_SEPARATOR, $subname, 2);
-            ErrorHandler::stop();
+            if (false !== strpos($subname, DIRECTORY_SEPARATOR)) {
+                list($entry, $subname) = explode(DIRECTORY_SEPARATOR, $subname, 2);
+            } else {
+                $entry   = $subname;
+                $subname = null;
+            }
+
             $currentFolder = $currentFolder->$entry;
+
             if (! $subname) {
                 break;
             }
--- a/src/Storage/Folder/Mbox.php
+++ b/src/Storage/Folder/Mbox.php
@@ -133,7 +133,7 @@ class Mbox extends Storage\Mbox implements FolderInterface
         $currentFolder = $this->rootFolder;
         $subname = trim($rootFolder, DIRECTORY_SEPARATOR);
         while ($currentFolder) {
-            ErrorHandler::start(E_NOTICE);
+            ErrorHandler::start(E_NOTICE | E_WARNING);
             list($entry, $subname) = explode(DIRECTORY_SEPARATOR, $subname, 2);
             ErrorHandler::stop();
             $currentFolder = $currentFolder->$entry;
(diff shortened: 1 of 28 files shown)


src/Storage/Folder/Mbox.php
```diff
@@ -125,7 +133,7 @@ public function getFolders($rootFolder = null)
         $currentFolder = $this->rootFolder;
         $subname = trim($rootFolder, DIRECTORY_SEPARATOR);
         while ($currentFolder) {
-            ErrorHandler::start(E_NOTICE);
+            ErrorHandler::start(E_NOTICE | E_WARNING);
```
Ocramius: An error handler used around an `explode()` call? O_o

src/Storage/ParamsNormalizerTrait.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Laminas\Mail\Storage;
+
+use Traversable;
+use Webmozart\Assert\Assert;
+
+trait ParamsNormalizerTrait
```
Ocramius: Should be `/** @internal */`

src/Storage/ParamsNormalizerTrait.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param mixed $params
+     * @return array<string, mixed>
+     */
+    private function normalizeParams($params): array
```
Ocramius: Let's not use a trait for this: a `public static function` works just as fine, without the complexities of traits

src/Storage/Folder/Mbox.php
Ocramius: Also, the added `E_WARNING` seems a bit much - perhaps we really need to just check `$subname` before calling `explode()` ?

src/Storage/Maildir.php
```diff
@@ -295,14 +301,14 @@ protected function getMaildirFiles($dh, $dirname, $defaultFlags = [])
                 continue;
             }
 
-            ErrorHandler::start(E_NOTICE);
+            ErrorHandler::start(E_NOTICE | E_WARNING);
```
Ocramius: Similar to above: a whole `ErrorHandler` usage just because we operate with `explode()` on unknown input (or assume it will produce less than 2 results)? :thinking:

src/Storage/ParamsNormalizerTrait.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        if ($params instanceof Traversable) {
```
Ocramius: This logic needs a unit test, once converted into a `static` method

Ocramius (changes requested):

Ocramius (approved):
LGTM 👍

PR #167, opened 2021-09-21
fix has timed out

### LINKED ISSUE #169: Error in AbstractProtocol.php prevents sending via SMTP
Hi there,
Since release 1.14.2 the timeout check is inverted leading to time out errors where the SMTP connection is functional.
See:

https://github.com/laminas/laminas-mail/blob/d219bfdc2677cbf89de9d1c622185f1c5f1f6fbe/src/Protocol/AbstractProtocol.php#L293

Please correct. Shall I make a PR? Not quite shure why

```
if (! empty($data['timed_out'])) {
```

was changet to

```
if (! $data['timed_out']) {
```

that would probably do, if `$data['time_out']` is always set, else revert it to the original...

```
if ($data['timed_out']) {
```

Best Regards,

Willi

### LINKED ISSUE #170: Laminas Mail 2.15 breaks SMTP in Magento 2.4.3
### Bug Report

This new version that fully breaks Magento 2.4.3 SMTP?? 

How can you guys release something like this that fully breaks Magento 2.4.3 SMTP?? I confirmed its this version by rolling back older version and now all mails send again fine.

#### Current behavior

Doimg the SMTP test in Admin this extension causes SMTP on port 465 SSL to time-out. Even though it times out it sends message 1 or 2 of the test emails. Trying to send regular emails of course all failed from contact us page to order confirm, etc.

#### How to reproduce

Try and do the test emails in M 2.4.3. I use Mailgun with the SMTP extension from MagePal. You will see from the test that it immediately times out with the error message. Still message 1 of 2 sends and arrives. Not 2of 2.

#### Expected behavior

Expect the test email to send both test emails and not deliver a timeout on port 465. Again, your older version was installed and works fine. Please fix, This could not have been tested with the latest M 2.4.3.

kokspflanze:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

If there is no timeout it goes into the RuntimeException, so its not possible to send mails.

### DIFF
--- a/src/Protocol/AbstractProtocol.php
+++ b/src/Protocol/AbstractProtocol.php
@@ -296,7 +296,7 @@ abstract class AbstractProtocol
         // Check meta data to ensure connection is still valid
         $info = stream_get_meta_data($this->socket);
 
-        if (! $info['timed_out']) {
+        if ($info['timed_out']) {
             throw new Exception\RuntimeException($this->host . ' has timed out');
         }
 
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "laminas/laminas-servicemanager": "^3.2.1",
         "phpunit/phpunit": "^9.3",
         "psalm/plugin-phpunit": "^0.15.1",
+        "symfony/process": "^5.3",
         "vimeo/psalm": "^4.7"
     },
     "suggest": {
--- a/composer.lock
+++ b/composer.lock
@@ -1457,43 +1457,46 @@
         },
         {
             "name": "laminas/laminas-servicemanager",
-            "version": "3.10.0",
+            "version": "3.7.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/laminas/laminas-servicemanager.git",
-                "reference": "e52b985909e0940bf22d34f322eb3f48bbef6bd1"
+                "reference": "2b0aee477fdbd3191af7c302b93dbc5fda0626f4"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/laminas/laminas-servicemanager/zipball/e52b985909e0940bf22d34f322eb3f48bbef6bd1",
-                "reference": "e52b985909e0940bf22d34f322eb3f48bbef6bd1",
+                "url": "https://api.github.com/repos/laminas/laminas-servicemanager/zipball/2b0aee477fdbd3191af7c302b93dbc5fda0626f4",
+                "reference": "2b0aee477fdbd3191af7c302b93dbc5fda0626f4",
                 "shasum": ""
             },
             "require": {
                 "container-interop/container-interop": "^1.2",
                 "laminas/laminas-stdlib": "^3.2.1",
-                "php": "~7.4.0 || ~8.0.0 || ~8.1.0",
+                "laminas/laminas-zendframework-bridge": "^1.0",
+                "php": "^7.3 || ~8.0.0",
                 "psr/container": "^1.0"
             },
             "conflict": {
                 "laminas/laminas-code": "<3.3.1",
-                "zendframework/zend-code": "<3.3.1",
-                "zendframework/zend-servicemanager": "*"
+                "zendframework/zend-code": "<3.3.1"
             },
             "provide": {
                 "container-interop/container-interop-implementation": "^1.2",
                 "psr/container-implementation": "^1.0"
             },
+            "replace": {
+                "zendframework/zend-servicemanager": "^3.4.0"
+            },
             "require-dev": {
                 "composer/package-versions-deprecated": "^1.0",
-                "laminas/laminas-coding-standard": "~2.2.1",
+                "laminas/laminas-coding-standard": "~2.2.0",
                 "laminas/laminas-container-config-test": "^0.3",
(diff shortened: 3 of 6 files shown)


src/Protocol/AbstractProtocol.php
```diff
@@ -296,7 +296,7 @@ protected function _receive($timeout = null)
         // Check meta data to ensure connection is still valid
         $info = stream_get_meta_data($this->socket);
 
-        if (! $info['timed_out']) {
```
Ocramius, line 299: Is there a meaningful way to verify this / avoid further regressions?

Ocramius (changes requested):

src/Protocol/AbstractProtocol.php
kokspflanze: Before the last release 2.14.2 it was `if (! empty($info['timed_out'])) {`.

I can also change it back to the old way.
Both ways would work fine for me.

Ocramius: I suppose `$info['timed_out']` can also be absent.

My question is whether there's a test case for this.

kokspflanze: from docs it looks like, its everytime there.
i also have no test case with a timed_out connection.

My case was only that all connections goes into that error, so no mailing was possible. (starting with 2.14.2).
For me looks like a simple true/false misstake.

So i created a fast easy PR, to help other ppl not waste again 1h to search that error.

Ocramius: > For me looks like a simple true/false misstake.

Yes, but while I understand the fix, we want tests to prevent this from happening again.

The question is: is this testable? Can we do it, to prevent the error from re-appearing in future changes/refactoring?

 > So i created a fast easy PR, to help other ppl not waste again 1h to search that error.

That's exactly what a test does :)

alexeyamitech:
This is affecting magento 2 module "mageplaza/magento-2-smtp"
The error message logged "smtp.gmail.com has timed out", which makes you think something is wrong with a server connectivity.

src/Protocol/AbstractProtocol.php
lcobucci: I didn't checkout the code but something like this **could** work (many warnings here because opening a connection with a random service and because the implementation is naive):

```php
/** @test */
public function exceptionShouldBeRaisedWhenConnectionHasTimedOut(): void
{
    $protocol = new class('httpstat.us', 80) extends AbstractProtocol {
        use ProtocolTrait;

        public function connect(): void
        {
            $this->socket = $this->setupSocket('tcp', $this->host, $this->port, 2);
        }

        public function send(string $path, int $readTimeout): string
        {
            $this->_send('GET ' . $path . ' HTTP/1.1');
            $this->_send('Host: ' . $this->host);
            $this->_send('');

            return $this->_read($readTimeout);
        }
    };
    
    $protocol->connect();
    self::assertSame('HTTP/1.1 200 OK' . AbstractProtocol::EOL, $protocol->send('/200', 10));

    $this->expectExceptionObject(new Laminas\Mail\Protocol\Exception\RuntimeException('httpstat.us has timed out'));
    $protocol->send('/200?sleep=2000', 1);
}
```

Maybe custom stream wrappers can help us to isolate things but it might be a bit trickier...

Ocramius:
@Slamdunk I suggest running a `composer update` after setting `config.platform.php=7.3.99` or such: then this could work 👍

Slamdunk:
I'm fixing CS/SA, but PHP 7.3 kicks on Githug Actions:
```
Symfony\Component\Process\Exception\ProcessTimedOutException: The process "'/usr/bin/php7.3' '-S' '127.0.0.1:8080' '-t' '/github/workspace/test/Protocol/HttpStatusService'" exceeded the timeout of 60 seconds.
```
May we ignore this error, or better, raise minimum PHP version?

Ocramius:
> raise minimum PHP version?

For a patch release we cannot, but we can skip the test on lower versions, and bump the dependencies on 2.15

src/Protocol/AbstractProtocol.php
Slamdunk: Thank you @lcobucci for this proto; I've made `httpstat.us` service local and the test is perfect.

This is what OSS should be: someone with good ideas, someone with time and will to implement them, everyone involved.

We need a new licence: get paid to collaborate, free to use, pay to complain

perfpcs:
Hello, So will all these changes also be added to M 2.4.3  with release of 2.15.x or next version?  I am just wondering how it will update to the newer files. Sure I could add manually but figure I'd ask here what the plan is for release to all?
thanks

froschdesign:
@perfpcs 
The releases are already there:

* https://github.com/laminas/laminas-mail/releases/tag/2.14.3
* https://github.com/laminas/laminas-mail/releases/tag/2.15.1

And both contain the bug fix.

Ocramius:
Thanks @Slamdunk! :+1:

perfpcs:
Awesome thank you @froschdesign  and @Slamdunk . You guys got on this fast!

froschdesign:
Thanks goes to @Slamdunk @lcobucci and @Ocramius!

PR #174, opened 2021-10-21
Fold a string with the correct length

danielabyan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

According to the RFC 2822 specification (clause 2.1.1), the header length should be no more than 78 characters.

> Each line of characters MUST be no more than 998 characters, and SHOULD be no more than 78 characters, excluding the CRLF.

According to the specification, the length of the ENTIRE LINE should not exceed 78 characters. Currently, the `laminas-mail` library only counts the length of a **value** of the header. 

### Reproducing the problem
```
$message = new Message();
$message->addFrom('matthew@example.org', 'Matthew Somelli');
$message->addTo('foobar@example.com');
$message->setSubject('xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx');
$message->setBody('This is the message body.');

echo $message->toString();
```

This code will return the following result.

> Date: Thu, 21 Oct 2021 09:59:33 +0000
> From: Matthew Somelli <matthew@example.org>
> To: foobar@example.com
> Subject: xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx
> 
> This is the message body.

As you can see from the example, the header is longer than 78 characters.

### New code behavior

This commit has fixed this issue. In this pool request, the code above will return the following result.

> Date: Thu, 21 Oct 2021 10:04:43 +0000
> From: Matthew Somelli <matthew@example.org>
> To: foobar@example.com
> Subject: xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx
>  xxxxx xxxxx
> 
> This is the message body.
> 

Now, the title does not exceed 78 characters.

### Note
Waiting for pooling request on [laminas-mime](https://github.com/laminas/laminas-mime/pull/25)

### DIFF
--- a/src/Header/HeaderWrap.php
+++ b/src/Header/HeaderWrap.php
@@ -12,9 +12,11 @@ use function iconv_mime_decode;
 use function iconv_mime_encode;
 use function implode;
 use function str_contains;
+use function str_pad;
 use function str_starts_with;
 use function strlen;
 use function strpos;
+use function substr;
 use function wordwrap;
 
 use const ICONV_MIME_DECODE_CONTINUE_ON_ERROR;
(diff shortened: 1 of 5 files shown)


src/Header/HeaderWrap.php
```diff
@@ -39,11 +39,23 @@ public static function wrap($value, HeaderInterface $header)
...
+            /*
+             * Before folding the header line, it is necessary to calculate the length of the
+             * entire header (including the name and colon). We need to put a stub at the
+             * beginning of the value so that the folding is performed correctly.
+             */
+            $headerLine = str_pad('0', $headerNameColonSize) . $value;
+            $foldedHeaderLine = wordwrap($headerLine, 78, Headers::FOLDING);
+
+            // Remove the stub and return the header folded value.
+            return mb_substr($foldedHeaderLine, $headerNameColonSize);
```
glensc: mb_substr must never be used in a library without a charset parameter. must not have side effects due different ini values in host environment.

also, you don't need mb_* function here at all.

danielabyan: Fixed

Slamdunk (approved):

PR #176, opened 2021-11-05
Drop `true/Punycode` in favor of `symfony/polyfill-intl-idn`

### LINKED ISSUE #175: ext-intl must be required in composer.json
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.15.1

#### Summary

I updated laminas/laminas-mail in a small application. Afterwards I could not send emails.

#### Current behavior

This component generates a fatal error:

```
Fatal error:  Uncaught Error: Undefined constant "Laminas\Mail\Header\IDNA_ERROR_EMPTY_LABEL" in /var/www/html/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php:295
Stack trace:
#0 /var/www/html/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php(135): Laminas\Mail\Header\AbstractAddressList->idnToAsciiViaPunycode('example.com')
#1 /var/www/html/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php(162): Laminas\Mail\Header\AbstractAddressList->idnToAscii('example.com')
#2 /var/www/html/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php(221): Laminas\Mail\Header\AbstractAddressList->getFieldValue(true)
#3 /var/www/html/vendor/laminas/laminas-mail/src/Headers.php(491): Laminas\Mail\Header\AbstractAddressList->toString()
#4 /var/www/html/public/test.php(11): Laminas\Mail\Headers->toString()
#5 {main}
  thrown in /var/www/html/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php on line 295
```

I think it was introduced in d75a9b29f3ad1bb6ab0e3c58e01bc5f48721e69c.

#### How to reproduce

```php
<?php
require_once 'vendor/autoload.php';

(Laminas\Mail\Headers::fromString('To: foo@example.com'))
    ->setEncoding('UTF-8')
    ->toString();
```

#### Expected behavior

Headers should be converted to strings w/o fatal error.

ghostwriter:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Drop `true/Punycode` in favor of `symfony/polyfill-intl-idn`

- Resolves #175
- Add failing test 
- Remove ext-intl extension for testing
- Remove missing intl extension constants and functions
- Drop `true/Punycode`
- Require `symfony/polyfill-intl-idn`
- Allow `composer/package-versions-deprecated` composer plugin
- Update composer.lock
- Update psalm-baseline.xml

### DIFF
--- a/.laminas-ci/pre-run.sh
+++ b/.laminas-ci/pre-run.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+INTL_EXTENSION=$(php -r 'echo sprintf("php%s.%s-intl",PHP_MAJOR_VERSION,PHP_MINOR_VERSION);')
+
+echo -e "Removing $INTL_EXTENSION extension:\n"
+sudo apt remove "$INTL_EXTENSION" -y
+
+echo -e "Cleaning up:\n"
+sudo apt autoclean && sudo apt autoremove
+
+echo -e "Installed extensions:\n"
+/usr/local/bin/php-extensions-with-version.php
--- a/src/Header/AbstractAddressList.php
+++ b/src/Header/AbstractAddressList.php
@@ -7,8 +7,6 @@ use Laminas\Mail\AddressList;
 use Laminas\Mail\Headers;
 use Laminas\Mail\Storage\Exception\RuntimeException;
 use Throwable;
-use TrueBV\Exception\OutOfBoundsException;
-use TrueBV\Punycode;
 
 /**
  * Base class for headers composing address lists (to, from, cc, bcc, reply-to)
@@ -52,11 +50,6 @@ abstract class AbstractAddressList implements HeaderInterface
      * @var string lower case field name
      */
     protected static $type;
-
-    /**
-     * @var Punycode|null
-     */
-    private static $punycode;
 
     public static function fromString($headerLine)
     {
@@ -128,11 +121,24 @@ abstract class AbstractAddressList implements HeaderInterface
      */
     protected function idnToAscii($domainName)
     {
-        if (function_exists('idn_to_ascii')) {
-            return $this->idnToAsciiViaIntl($domainName);
+        $ascii = idn_to_ascii($domainName, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46, $conversionInfo);
+        if (false !== $ascii) {
+            return $ascii;
         }
 
-        return $this->idnToAsciiViaPunycode($domainName);
+        $messages = [];
+        $errors   = (int) $conversionInfo['errors'];
+
+        foreach (self::IDNA_ERROR_MAP as $flag => $message) {
+            if (($flag & $errors) === $flag) {
+                $messages[] = $message;
+            }
+        }
+
+        throw new RuntimeException(sprintf(
+            'Failed encoding domain due to errors: %s',
+            implode(', ', $messages)
+        ));
     }
 
     public function getFieldValue($format = HeaderInterface::FORMAT_RAW)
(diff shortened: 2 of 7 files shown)


.laminas-ci/pre-run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+INTL_EXTENSION=$(php -r 'echo sprintf("php%s.%s-intl",PHP_MAJOR_VERSION,PHP_MINOR_VERSION);')
+
+echo "Removeing $INTL_EXTENSION extension:"
```
glensc:
```suggestion
echo "Removing $INTL_EXTENSION extension:"
```

.laminas-ci/pre-run.sh
```diff
@@ -0,0 +1,14 @@
...
+sudo apt remove "$INTL_EXTENSION" -y
+
+echo "Cleaning up:"
+sudo apt autoclean && sudo apt autoremove
+
+echo "Installed extensions:"
+/usr/local/bin/php-extensions-with-version.php
```
glensc, line 14: missing EOL

src/Header/AbstractAddressList.php
```diff
@@ -10,6 +10,24 @@
 use TrueBV\Exception\OutOfBoundsException;
 use TrueBV\Punycode;
 
+if (! extension_loaded('intl')) {
+    define('IDNA_DEFAULT', 0);
```
glensc: i'm sure this is not okay test, what if the `symfony/polyfill-intl-idn polyfill` is loaded (or any other), in that case the `extension_loaded('intl')` is `false`, but the constants are defined.

you should check each constant on it's own for such cases

glensc: also, perhaps this is not the way you should recover this. the errors came because the constants are being defined as class constant `self::IDNA_ERROR_MAP`:
- https://github.com/laminas/laminas-mail/blob/2.15.1/src/Header/AbstractAddressList.php#L18-L32

but move these to some function, `getIdnaErrorMap()`, so they would be referenced only if using ext-intl mode, which means you don't need to affect PHP runtime global state by defining constants

test/HeadersTest.php
```diff
@@ -647,4 +647,14 @@ public function testStrictKeyComparisonInGet(): void
         $headers = Mail\Headers::fromString("000: foo-bar");
         $this->assertFalse($headers->get('0'));
     }
+
+    /** @group issues-175 */
```
glensc:
```suggestion
    /** @group issue-175 */
```

makes more sense to use singular than plural

src/Header/AbstractAddressList.php
glensc: Maybe an even better refactor would be to extract the idn/intl stuff to own class, out of `AbstractAddressList`, due to S of SOLID principle, IdnaConverter->idnToAscii perhaps?
- https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design

ps, there's another idn related library:
- https://github.com/algo26-matthias/idna-convert

ghostwriter: Thanks for the feedback, I'll revisit this section tomorrow.

ghostwriter: thoughts on requiring [`symfony/polyfill-intl-idn`](https://github.com/symfony/polyfill-intl-idn) via composer instead?

glensc: but, look, the other idn library is already in the required library list:
- https://github.com/laminas/laminas-mail/blob/2.15.1/composer.json#L18

why you getting intl errors is because of bad implementation (using constants of library that's not used), that's the root of the problem.

src/Header/AbstractAddressList.php
```diff
@@ -128,11 +113,19 @@ public function getFieldName()
      */
     protected function idnToAscii($domainName)
     {
-        if (function_exists('idn_to_ascii')) {
-            return $this->idnToAsciiViaIntl($domainName);
+        if (null === self::$punycode) {
+            self::$punycode = new ToIdn();
```
Slamdunk: DocBlock type for `self::$punycode` must be updated as well.

Slamdunk (changes requested):
Overall looks good to me.

Please try to raise the minimum version of deps to see if build passes, and check out the few errors in both static analysis and coding standard :muscle:

Slamdunk (approved):

Slamdunk:
Thank you @ghostwriter

PR #183, opened 2022-02-22
Trim surrounding single quotes in emails to support Outlook-style single-quotes around `To:` header weirdness

CapnFelix:
Outlook does add single quotes to emails sometimes which is not valid

|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | yes
| BC Break      |no
| New Feature   |no
| RFC           | no
| QA            | no

### Description

We're seeing emails generated by outlook which do have surrounding single quote like this:
`To: <'user@example.com'>`
The fix does just trim those.

Issue on microsoft.com: https://answers.microsoft.com/en-us/outlook_com/forum/all/single-quote-marks-around-email-addresses-in/37122680-0ea2-46ed-bfeb-8f67d3441e5f
 
This is my first PR, so please let me know if did something wrong :)

### DIFF
--- a/src/Address.php
+++ b/src/Address.php
@@ -44,6 +44,8 @@ class Address implements Address\AddressInterface
             $email = $matches['email'];
         }
         $email = trim($email);
+        //trim single quotes, because outlook does add single quotes to emails sometimes which is technically not valid
+        $email = trim($email, '\'');
 
         return new static($email, $name, $comment);
     }
--- a/test/_files/mail_surrounding_single_quotes.eml
+++ b/test/_files/mail_surrounding_single_quotes.eml
@@ -1,0 +1,27 @@
+To: <'foo@example.com'>
+Subject: multipart
+Date: Sun, 01 Jan 2000 00:00:00 +0000
+From: =?UTF-8?Q?"Peter M=C3=BCller"?= <peter-mueller@example.com>
+ContENT-type: multipart/alternative; boUNDary="crazy-multipart"
+Message-ID: <CALTvGe4_oYgf9WsYgauv7qXh2-6=KbPLExmJNG7fCs9B=1nOYg@mail.example.com>
+MIME-version: 1.0
+
+multipart message
+--crazy-multipart
+Content-type: text/plain
+
+The first part 
+is horizontal
+
+--crazy-multipart
+Content-type: text/x-vertical
+
+T s p i v
+h e a s e
+e c r   r
+  o t   t
+  n     i
+  d     c
+        a
+        l
+--crazy-multipart--
--- a/test/Header/AddressListHeaderTest.php
+++ b/test/Header/AddressListHeaderTest.php
@@ -161,6 +161,25 @@ class AddressListHeaderTest extends TestCase
             ['From: (Comment\\\\)user@example.com(Another)'],
         ];
     }
+    /**
+     * @dataProvider getHeadersWithSurroundingSingleQuotes
+     */
+    public function testTrimSurroundingSingleQuotes($value): void
+    {
+        $header = To::fromString($value);
+        $list = $header->getAddressList();
+        $this->assertEquals(1, count($list));
+        $this->assertTrue($list->has('foo@example.com'));
+    }
+
+    public function getHeadersWithSurroundingSingleQuotes(): array
+    {
+        return [
+            ['To: <\'foo@example.com\'>'],
+            ['To: Foo Bar <\'foo@example.com\'>'],
+            ['To: \'foo@example.com\''],
+        ];
+    }
 
     /**
      * @group 3789
--- a/test/Header/AddressListHeaderTest.php
+++ b/test/Header/AddressListHeaderTest.php
@@ -161,10 +161,11 @@ class AddressListHeaderTest extends TestCase
             ['From: (Comment\\\\)user@example.com(Another)'],
         ];
     }
+
     /**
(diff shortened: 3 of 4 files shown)


src/Address.php
```diff
@@ -44,6 +44,8 @@ public static function fromString($address, $comment = null)
             $email = $matches['email'];
         }
         $email = trim($email);
+        //trim single quotes, because outlook does add single quotes to emails sometimes which is technically not valid
+        $email = trim($email, '\'');
```
Ocramius, line 48: This should already fail validation above, no?

Ocramius (changes requested):
Tests strictly required to validate such a change: IMO not valid to accept a string within quotes as a valid `Address` anyway, as such kind of cleanup should occur wherever you input data into the system that uses `Address::fromString()`

src/Address.php
CapnFelix: Hm, no... the exception is actually thrown in the __contsruct():

```
Laminas\Mail\Exception\InvalidArgumentException: ‘example.com'' is not a valid hostname for the email address in /var/www/html/projectname/vendor/laminas/laminas-mail/src/Address.php:72
Stack trace:
#0 /var/www/html/projectname/vendor/laminas/laminas-mail/src/Address.php(48): Laminas\Mail\Address->__construct(‘’email…’, NULL, '')
#1 /var/www/html/projectname/vendor/laminas/laminas-mail/src/Header/AbstractAddressList.php(99): Laminas\Mail\Address::fromString('__construct(Array)
#11 /var/www/html/projectname/vendor/laminas/laminas-mail/src/Storage/AbstractStorage.php(270): Laminas\Mail\Storage\Imap->getMessage(3)
#12 /var/www/html/projectname/code/MyLaminasImapClient.php(137): Laminas\Mail\Storage\AbstractStorage->current() 
…
```

CapnFelix:
Thanks for the quick feedback.
I agree that stripping it in ::fromString is not perfect, even though when looking at the caller stack, I'm not quite sure where to put this instead.

Message is loaded from IMAP in
`Laminas\Mail\Storage\Imap->getMessage()`
which is then calling `\Laminas\Mail\Storage\Message`
and then finally using `\Laminas\Mail\Header\AbstractAddressList::fromString` to generate the AddressList.
I think it would be possible to do the trimming here, but I would need to use the same reg ex to unpack the email address.
What do you think?

Ocramius:
Aha, the fact that it is loaded via IMAP is new info.

I wonder if it is possible to get one of the loaded raw messages as part of data provider entries for our test suite? That would make it clear that we're talking about a real-world IMAP use-case, and prevent future regressions.

CapnFelix:
I am sorry, I should have provided this info in the initial description.
I've added a regression test to verify the change + added an example mail to the files folder.
Does this make sense?

test/_files/mail_surrounding_single_quotes.eml
```diff
@@ -0,0 +1,27 @@
+To: <'foo@example.com'>
```
Ocramius, line 1: We should probably use this mail in our tests somewhere :D

CapnFelix: makes sense... :D 
haven't found the place so far, will look into that later...

Ocramius (changes requested):
Tests make sense to me, although the type errors reported by psalm should really be checked (see them on the diff).

As for the `trim()` call, I'm thinking that it should be moved into `Message`, rather than in `Address`. Rationale: `Address::fromString()` is used outside the context of `Message` too, while it looks like this Outlook IMAP oddity is only applicable when parsing messages in Outlook.

CapnFelix:
Maybe there is a small misunderstanding: We're looking at a message sent from an outlook client to a Gmail inbox. We're actually downloading this message via IMAP from Gmail in this case. 
So I guess this is more a general problem not related to the storage since it is the email that is "wrong" no matter which protocol I'm using to download it.
But I can look into a way to do that fix in `Message`...

CapnFelix:
Some thoughts looking at it a little more in depths:
what the `Address::fromString` really does is extracting name and email from an address header value string. So I think it would make sense to enhance the extracting to be able to deal with the single quotes either by adjusting the preg_match or by doing the trimming.
When looking at the `Message` class I don't really have a good idea how to fix it there.
Let me know what you think! 
Thanks!

Ocramius:
Reasoning seems fine: are you able to bring CI to green, so that somebody else can throw in a more informed review on whether putting this responsibility in `Address` is acceptable?

Slamdunk:
> Issue on microsoft.com: https://answers.microsoft.com/en-us/outlook_com/forum/all/single-quote-marks-around-email-addresses-in/37122680-0ea2-46ed-bfeb-8f67d3441e5f

I've read https://datatracker.ietf.org/doc/html/rfc5322 twice, and nowhere it allows `<'foo@example.com'>` as valid syntax :facepalm: 

But as maintainer of [ddeboer/imap](https://github.com/ddeboer/imap) it's no surprise to me

Slamdunk (approved):

Ocramius:
Ok, so let's get it green, then merged

CapnFelix:
Awesome! 
I am not familiar with psalm yet, but I think I got it to work :D

Ocramius (approved):
Awesome! :ship: 

Thanks @CapnFelix! Releasing nao.

CapnFelix:
Thank you so much! 🌊

PR #191, opened 2022-07-13
Adjust documentation to avoid accumulating recipients in a loop (resulting in multiple mails received by same recipient)

### LINKED ISSUE #189: Doc sending-multiple-mails example error
### Bug Report

Documentation issue

#### Summary

The example https://docs.laminas.dev/laminas-mail/transport/smtp-multiple-send/#sending-multiple-mails-per-smtp-connection_1
Should replace recipient instead of accumulate them in the "To" header.

#### Current behavior
It adds recipient.

**`$message->addTo($address);`**

So, if 10 recipients, the first recipient, for example, will receive 10 times the email.

```
foreach ($recipients as $address) {
    // Clone the message and add a recipient:
    $message = clone $template;
    $message->addTo($address);

    $transport->send($message);
}
```

#### Expected behavior
Replacing recipient.

**`$message->setTo($address);`**

```
foreach ($recipients as $address) {
    // Clone the message and add a recipient:
    $message = clone $template;
    $message->setTo($address);

    $transport->send($message);
}
```
#### About cloning
In this example cloning is not useful.

So, the best and simple example is:

```
use Laminas\Mail\Message;
use Laminas\Mail\Transport\Smtp;

// Create transport
$transport = new Smtp([
    'host' => 'mail.example.com'
]);

// Create a base message:
$message = (new Message()) // to doc reviewer : since PHP 5.4
                ->addFrom('sender@example.com', 'John Doe')
                ->addReplyTo('replyto@example.com', 'Jane Doe'
                ->setSubject('Demo of multiple mails per SMTP connection')
                ->setBody('... Your message here ...');

// Loop through recipients:
foreach ($recipients as $address) {
    $message->setTo($address);
    $transport->send($message);
}
```

bouks:
Provide a better example with changing of recipient instead of accumulating them and sending multiple times to each recipient.
Add better message instantiation and usage syntax (since PHP 5.4 !)
Remove cloning, that is unnecessary.

Fixes #189 

Signed-off-by: bouks <karibou@ribouka.com>

|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/docs/book/transport/smtp-multiple-send.md
+++ b/docs/book/transport/smtp-multiple-send.md
@@ -19,18 +19,15 @@ $transport = new Smtp([
 ]);
 
 // Create a base message:
-$template = new Message();
-$template->addFrom('sender@example.com', 'John Doe');
-$template->addReplyTo('replyto@example.com', 'Jane Doe');
-$template->setSubject('Demo of multiple mails per SMTP connection');
-$template->setBody('... Your message here ...');
+$message = (new Message())
+                ->addFrom('sender@example.com', 'John Doe')
+                ->addReplyTo('replyto@example.com', 'Jane Doe')
+                ->setSubject('Demo of multiple mails per SMTP connection')
+                ->setBody('... Your message here ...');
 
 // Loop through recipients:
 foreach ($recipients as $address) {
-    // Clone the message and add a recipient:
-    $message = clone $template;
-    $message->addTo($address);
-
+    $message->setTo($address);
     $transport->send($message);
 }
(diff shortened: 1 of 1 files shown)


Ocramius:
@bouks I think you meant to push this to your `patch-1` branch?

You should be able to edit that from the Github UI :thinking:

Ocramius (approved):
Thanks @bouks!

PR #199, opened 2022-07-28
POP3 XOAUTH2 support for Microsoft

griesi007:
Add support for Microsoft POP3 via modern authentication

Usage:
`$storage = new \Laminas\Mail\Storage\Pop3Pop3(\Laminas\Mail\Protocol\Pop3\Xoauth2\Microsoft::fromParams([
	'host' => 'Outlook.office365.com',
	'targetMailbox' => 'test@contoso.onmicrosoft.com',
	'accessToken' => '###ACCÈSSTOKEN###',
]));
`

PR #200, opened 2022-07-29
Add support for Microsoft POP3 XOAUTH2

griesi007:
Signed-off-by: griesi007 <noreply@12live.de>

Add support for Microsoft POP3 via modern authentication in addition to rejected pull request #199 with required integration tests

```php
$storage = new \Laminas\Mail\Storage\Pop3(
	Laminas\Mail\Protocol\Pop3\Xoauth2\MicrosoftFactory::getInstance(
		'Outlook.office365.com',
		'mail@example.com',
		'accessToken'
	)
);
```


src/Protocol/Pop3/Xoauth2/Microsoft.php
```diff
@@ -0,0 +1,31 @@
...
+    protected const AUTH_INITIALIZE_REQUEST = 'AUTH XOAUTH2';
+    protected const AUTH_RESPONSE_INITIALIZED_OK = '+';
+
+    /**
+     * xoauth2Sasl - XOUATH2 encoded client response
+     * @param string $xoauth2Sasl
+     * @return void
+     */
+
+    public function authenticate(string $xoauth2Sasl): void
```
Slamdunk: Instead of adding a new API, I think we can reuse the `login` method.

src/Protocol/Pop3/Xoauth2/MicrosoftFactory.php
```diff
@@ -0,0 +1,42 @@
+<?php
+declare(strict_types=1);
+
+namespace Laminas\Mail\Protocol\Pop3\Xoauth2;
+
+use Laminas\Mail\Protocol\Xoauth2\Xoauth2;
+
+class MicrosoftFactory
```
Slamdunk: I don't think we need a factory for this.

The user could will look like:

```php
$protocol = new Xoauth2\Microsoft("host", "port", "ssl");
$protocol->login("user", "token");

$storage = new Pop3($protocol);
```

Users can do that by their own.

src/Protocol/Xoauth2/Xoauth2.php
```diff
@@ -0,0 +1,25 @@
+<?php
+declare(strict_types=1);
+
+
+namespace Laminas\Mail\Protocol\Xoauth2;
+
+class Xoauth2
```
Slamdunk: We are going to need this functionality in multiple part for sure, but we can make it `final` and `@internal`

src/Protocol/Pop3/Xoauth2/Microsoft.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Laminas\Mail\Protocol\Pop3\Xoauth2;
+
+use Laminas\Mail\Protocol\Exception\RuntimeException;
+use Laminas\Mail\Storage\ParamsNormalizer;
+
+class Microsoft extends \Laminas\Mail\Protocol\Pop3
```
Slamdunk: I see you didn't make it `final` because it needs to be tested with a subclass.
I'm ok with it (our fault architecture doesn't allow a better mocking experience) but still we can add a `@final` docblock, so inheritance in user code gets disabled.

test/Protocol/Pop3/Xoauth2/MicrosoftTest.php
```diff
@@ -0,0 +1,57 @@
...
+class MicrosoftTest extends TestCase
+{
+    public function testIntegration():void
+    {
+        $protocol = new class() extends Microsoft {
+
+            /** @var string $step */
+            private $step = '';
+
+            public function __construct()
```
Slamdunk: This can be removed

test/Protocol/Pop3/Xoauth2/MicrosoftTest.php
```diff
@@ -0,0 +1,57 @@
...
+            {
+                $this->socket = fopen("php://memory", 'rw+');
+            }
+        };
+
+        $protocol->connect('localhost', 0, false);
+
+        $protocol->authenticate(Xoauth2::encodeXoauth2Sasl('test@example.com', '123'));
+
+        $this->assertInstanceOf(Microsoft::class, $protocol);
```
Slamdunk, line 85: I'd like to see an assertion on the `socket`, that is has received the `AUTH XOAUTH2` string

Slamdunk (changes requested):
Thank you @griesi007 for the improvements over the first PR.
Check the few changes I pinpointed and let's see how the shape looks.
Please rebase the PR as the first step.

src/Protocol/Pop3/Xoauth2/Microsoft.php
griesi007: @Slamdunk yes we might. What I do not like with this idea is that we are forced to use the misleading parameters named username / password and the 3rd parameter which doesn´t have any effect at all when we opt-in for overwriting the login method. Basically it depends if you want to make the code as explicit as possible or as easy as possible to use. I would prefer the first approach.

Slamdunk: The third parameter can be disabled by rewriting the DocBlock comment in the `Pop3` sublcass.
The second parameter, the `password` one, can in fact be misleading, but we can write a good DocBlock comment for that too.

Explicitness is a good goal, but keep in mind that:

1. adding a new API represents a huge maintenance cost
2. the `login` API doesn't disappear: this class still has it, and it would be very confusing for the end user to have both API available

@Ocramius WDYT?

griesi007: @Slamdunk Another issue I see with the login method are missing type declarations for the params

griesi007:
@Slamdunk I am a bit confused about changes lately in Protocol\Pop3. composer.json specifies that laminas mail should be compatible to PHP7.3 but these changes aren´t "Catching exceptions without capturing them to variables ". Which is minimal required compatibility in laminas-mail?

src/Protocol/Pop3/Xoauth2/Microsoft.php
Slamdunk: This is still pending

griesi007:
@Slamdunk : what´s the current status on this pull request? All requested changes habe been applied some time ago. Is it still considered to be accepted and merged?

Slamdunk:
I'm sorry for the late reply, indeed I forgot this.

I don't know how to handle this properly though: I am not able to check that this PR works, I don't have an Office365 account.

@Ocramius @weierophinney beside the formal correctness of the code and adherence to `laminas` standards, is trusting the contributor enough to merge a PR with a new feature, when I cannot test it and its test cannot be automated?

Ocramius:
@Slamdunk if you are happy with just the unit test, then let's move it forward.

PR #201, opened 2022-08-01
Allow symfony/process v6 in dev dependencies

Slamdunk:


composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "4b81ea029125cc7fb1bd010d47c2cb6b",
+    "content-hash": "5285af5a71d6a09501993d6e09046422",
```
Ocramius, line 7: @Slamdunk this didn't change the `symfony/process` version that is installed - seems wrong?

I'd rather say that we need to drop `symfony/process` from `require-dev` completely: we don't use it, do we?

Slamdunk: > this didn't change the `symfony/process` version that is installed

Yes it did:

1. `v5.4.11` on 7.3 https://github.com/laminas/laminas-mail/runs/7612998749?check_suite_focus=true#step:3:302
2. `v5.4.11` on 7.4 https://github.com/laminas/laminas-mail/runs/7612999087?check_suite_focus=true#step:3:305
3. `v6.0.11` on 8.0 https://github.com/laminas/laminas-mail/runs/7612999512?check_suite_focus=true#step:3:308
4. `v6.1.3` on 8.1 https://github.com/laminas/laminas-mail/runs/7612999881?check_suite_focus=true#step:3:308

> we don't use it, do we?

Yes we do: https://github.com/laminas/laminas-mail/blob/2ec0498fd03dc4684e24b7c7e0f7d015ebabe6d5/test/Protocol/AbstractProtocolTest.php#L53

Ocramius: Thanks for checking that we test it! :muscle:

All good here then!

PR #202, opened 2022-08-02
Update laminas-coding-standard to v2

Slamdunk:


Slamdunk:
```
PHP Fatal error:  Uncaught TypeError: Argument 5 passed to SlevomatCodingStandard\Helpers\Annotation\ParameterAnnotation::__construct() must be an instance of PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode or null, instance of PHPStan\PhpDocParser\Ast\PhpDoc\TypelessParamTagValueNode given, called in /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationHelper.php on line 357 and defined in /github/workspace/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/Annotation/ParameterAnnotation.php:31
```
:sob:

Ocramius:
Yup, you can force `phpstan/phpdoc-parser: <1.6` to make it run, meanwhile (do it locally, run a ` composer update`, then remove the clause and use `composer update nothing`)

Slamdunk:
Wouldn't CI fail nevertheless, if I don't commit the `phpdoc-parser: <1.6` requirement?

Ocramius:
Only in the next lockfile update PR

Slamdunk:
As I expected, build can't get green if I don't commit `"phpstan/phpdoc-parser": "~1.5.0"`:
https://github.com/laminas/laminas-mail/runs/7627796218?check_suite_focus=true#step:3:416

I can commit it though, can't I?

Ocramius:
The idea is to commit a `composer.lock` with an older `phpstan/phpdoc-parser`, by leaving `composer.json` intact.
CS checks run on locked dependencies only.

Slamdunk:
Oh, now I get it.

But I don't see a laminas guide on how to produce the `composer.lock` consistently: is a `composer update` on lowest supported PHP binary enough?

Ocramius:
There's no guide for that, afaik: I proposed an approach in https://github.com/laminas/laminas-mail/pull/202#issuecomment-1202127637

 1. add the dependency restriction to `composer.json`
 2. run `composer update`
 3. remove the dependency restriction from `composer.json`
 4. run `composer update nothing`

this gives you a `composer.lock` in the state you need it.

Slamdunk:
Thank you, I was missing the last step.

Now, down to the PR: LCSv2 introduces `declare(strict_types=1);` everywhere, should this change be considered a BC Break?

Ocramius:
It really depends - `declare(strict_types=1)` is safe to perform in some cases, while in others, it can lead to massive breakages.

I wouldn't say that the `laminas/laminas-mail` code is very "safe", at least when looking at its large psalm baseline:

https://github.com/laminas/laminas-mail/blob/2ec0498fd03dc4684e24b7c7e0f7d015ebabe6d5/psalm-baseline.xml#L3904-L3908

I think we need to disable that rule here. See how it's done in `laminas/laminas-eventmanager`: 

https://github.com/laminas/laminas-eventmanager/blob/6c056ab3b4b464daa308fd9ca3e7416ca40cf200/phpcs.xml#L23-L25

src/Exception/ExceptionInterface.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Laminas\Mail\Exception;
 
-interface ExceptionInterface
+use Throwable;
+
+interface ExceptionInterface extends Throwable
```
Slamdunk, line 7: This is the main relevant change in the Psalm fixing process, it should be BC as far as I know

src/Header/AbstractAddressList.php
```diff
@@ -171,7 +171,7 @@ protected function idnToAscii($domainName): string
     }
 
     /**
-     * @param string $format
+     * @param bool $format
```
Ocramius: Wait, the format is not a string? :O

Ocramius: One thing we could improve here is documenting `@param HeaderInterface::FORMAT_* $format`, but non-blocker.

src/Header/Subject.php
```diff
@@ -115,8 +115,6 @@ public function getEncoding()
      */
     public function setSubject($subject)
     {
-        $subject = (string) $subject;
```
Ocramius, line 95: We still need this, even if redundant, for any people passing a non-string `$subject`

src/Protocol/Imap.php
```diff
@@ -747,7 +747,7 @@ public function append($folder, $message, $flags = null, $date = null)
      */
     public function copy($folder, $from, $to = null)
     {
-        $set = (int) $from;
+        $set = (string) $from;
```
Ocramius, line 750: This change is a bit cryptic :thinking:

src/Protocol/SmtpPluginManager.php
```diff
@@ -25,7 +25,7 @@ class SmtpPluginManager extends AbstractPluginManager
     /**
      * Service aliases
      *
-     * @var array
+     * @var array<array-key, class-string>
```
Ocramius, line 36: The template parameter on `AbstractPluginManager<InstanceType>` should probably be filled:

https://github.com/laminas/laminas-servicemanager/blob/6cee2f3125e074c0960c13a9ffc1ed93127b860e/src/AbstractPluginManager.php#L35

Slamdunk (commented):
I've carefully fixed all the PHPCS issues.

I've also fixed most of the notable PSalm issues; the rest that I've baselined are just too tedious to be done manually now in PHPDocs.

Similarly to what I've done for [`laminas-form:v3.0.0`](https://github.com/laminas/laminas-form/pulls?q=is%3Apr+author%3ASlamdunk+is%3Aclosed+milestone%3A3.0.0) I'll be happy to clear all PSalm baseline once we can ship native types though.

src/Storage/Folder/FolderInterface.php
```diff
@@ -3,14 +3,15 @@
...
+use Laminas\Mail\Storage\Folder;
 
 interface FolderInterface
 {
     /**
      * get root folder or given folder
      *
      * @param string $rootFolder get folder structure for given folder, else root
-     * @return FolderInterface root or wanted folder
+     * @return Folder root or wanted folder
```
Ocramius, line 14: Shouldn't this return `FolderInterface`? Why is the concrete instance returned from the interface?

src/Storage/Part.php
```diff
@@ -310,7 +310,7 @@ public function getHeaders()
      * @param  string $name   name of header, matches case-insensitive, but camel-case is replaced with dashes
      * @param  string $format change type of return value to 'string' or 'array'
      * @throws Exception\InvalidArgumentException
-     * @return string|array|HeaderInterface|ArrayIterator value of header in wanted or internal format
+     * @return string|true|array|HeaderInterface|ArrayIterator value of header in wanted or internal format
```
Ocramius: Hmm, `true`? :thinking:

Ocramius (changes requested):

src/Storage/Folder/FolderInterface.php
Slamdunk: This puzzled me too: `FolderInterface` represents a Storage contract (Imap, Mailbox, Maildir) while `Folder` is an actual folder in any of those storage.

src/Storage/Part.php
Slamdunk: I didn't follow the rabbit down the hole enough: proper PHPDoc fixing coming

src/Protocol/Imap.php
Slamdunk: `$from` already is an `int`, and it needs to be used as a `string` to compose the proper Imap raw request.

I have no idea why it was the other way, so I guess it was just a typo

src/Header/Subject.php
Slamdunk: :+1:

src/Header/AbstractAddressList.php
Slamdunk: > Wait, the format is not a string? :O

Heh :facepalm: 

> `@param HeaderInterface::FORMAT_*`

:+1:

src/Protocol/SmtpPluginManager.php
Slamdunk: I've looked at https://github.com/laminas/laminas-validator/pull/129 and commit https://github.com/laminas/laminas-mail/pull/202/commits/d20feb23f9e3806725e6e2d8dc2531cc5c09086c is the result

Slamdunk:
```
Error: src/Protocol/SmtpPluginManager.php:26:23: InvalidTypeImport: Type alias FactoriesConfigurationType imported from Laminas\ServiceManager\ConfigInterface
```

Because I've checked commit validity against latest deps, but PSalm in CI runs on lowest which misses all the latest types-related releases.

Should I revert this, or keep and baseline it?

src/Exception/ExceptionInterface.php
Ocramius: It's not BC, but then `ExceptionInterface` here is only designed to be **caught**, not **implemented**, so we're OK.

src/Header/Subject.php
Ocramius: 35aac10461a138f5e53115451857de2bca4cb439

src/Header/AbstractAddressList.php
Ocramius: e5b23be47d2bb87376e3ddd9656885d33b6decff

src/Protocol/Imap.php
Ocramius: Works for me, thanks!

src/Protocol/SmtpPluginManager.php
Ocramius: >  but PSalm in CI runs on lowest which misses all the latest types-related releases.

It should run with **locked** dependencies. 

This seems to occur because we're still on PHP 7.3, and the latest type improvements are on PHP 7.4.

Let's touch this in a follow-up that drops PHP 7.4 :+1:

src/Storage/Folder/FolderInterface.php
Ocramius: Thanks for clarifying! :+1:

Ocramius: Related: 88d54cb3d12496a23e731c66e438e09af524123d

src/Protocol/SmtpPluginManager.php
Ocramius: Ah, I see that you did add the templated types. In order for that to work, we need to drop PHP 7.3, and then a `composer update` **should** bring us on par with the servicemanager improvements, as well as make the build green again  :+1:

src/Storage/Part.php
Ocramius: 6e6ba8c0819e7aed57a03645306c0df2802d3b1d

Ocramius (approved):
LGTM.

What's still missing:

 * [ ] drop PHP 7.4
 * [ ] re-generate baseline with the new dependency upgrades (it should get rid of the templated types issues)

Slamdunk:
> What's still missing

Since you approved this PR, I guess I can merge this as-is (with failing build) and then open a follow up to drop PHP < 8.0.

Merging red builds isn't good, but neither is discussing such big changes in the same PR.
At least that's what I'd do in my repos.

Confirm please, just so that I can learn your preferred way of managing such situations.

Ocramius:
No, please make it green first, be it through adjusted baseline or else, but let's not merge on red, as that just blocks other work, and will also lead to raised PRs by Renovate bot overnight :|

PR #203, opened 2022-08-04
Drop PHP 7.3 and PHP 7.4 support

Slamdunk:


composer.json
```diff
@@ -8,32 +8,29 @@
...
+        "laminas/laminas-loader": "^2.8.0",
         "laminas/laminas-mime": "^2.9.1",
-        "laminas/laminas-stdlib": "^3.6",
-        "laminas/laminas-validator": "^2.15",
-        "symfony/polyfill-mbstring": "^1.12.0",
-        "webmozart/assert": "^1.10",
-        "symfony/polyfill-intl-idn": "^1.24.0"
-    },
-    "conflict": {
-        "zendframework/zend-mail": "*"
```
Slamdunk, line 22: https://github.com/zendframework/zend-mail/blob/release-2.10.0/composer.json#L19

Since PHP versions now are going to mismatch completely, we can drop this, can't we?

src/ConfigProvider.php
```diff
@@ -26,10 +24,6 @@ public function __invoke()
     public function getDependencyConfig()
     {
         return [
-            // Legacy Zend Framework aliases
-            'aliases'   => [
-                SmtpPluginManager::class => Protocol\SmtpPluginManager::class,
```
Ocramius: These should stay for BC, since it may still be referenced by string

src/Protocol/SmtpPluginManager.php
```diff
@@ -48,18 +44,6 @@ class SmtpPluginManager extends AbstractPluginManager
...
-        Crammd5::class                  => Smtp\Auth\Crammd5::class,
-        Login::class                    => Smtp\Auth\Login::class,
-        Plain::class                    => Smtp\Auth\Plain::class,
-        \Zend\Mail\Protocol\Smtp::class => Smtp::class,
-
-        // v2 normalized FQCNs
-        'zendmailprotocolsmtpauthcrammd5' => Smtp\Auth\Crammd5::class,
-        'zendmailprotocolsmtpauthlogin'   => Smtp\Auth\Login::class,
-        'zendmailprotocolsmtpauthplain'   => Smtp\Auth\Plain::class,
-        'zendmailprotocolsmtp'            => Smtp::class,
```
Ocramius: These should stay until the next major

src/Protocol/SmtpPluginManagerFactory.php
```diff
@@ -2,25 +2,33 @@
 
 namespace Laminas\Mail\Protocol;
 
-use Interop\Container\ContainerInterface;
+use interop\container\containerinterface;
```
Ocramius: This replacement is wrong due to PHPCS not recognizing `class_alias()`

test/Protocol/SmtpPluginManagerFactoryTest.php
```diff
@@ -2,7 +2,7 @@
 
 namespace LaminasTest\Mail\Protocol;
 
-use Interop\Container\ContainerInterface;
+use interop\container\containerinterface;
```
Ocramius: This replacement is wrong due to PHPCS not recognizing `class_alias()`

Ocramius (changes requested):

Ocramius (approved):
:ship: thanks @Slamdunk!

Ocramius:
@samsonasik I'm gonna hold off a release, if you want to try PHP 8 rector transformations here :-)

weierophinney:
What was the justification for dropping support for 7.4? Considering it's still receiving security patches, I'm not seeing why we'd drop it in a new minor until later in the year, unless there's some compatibility issue. If there is, it'd be nice to at least have a link so that others coming to the issue know what prompted the decision...

Slamdunk:
A cascade of dev dep updates lead us to choose either baseline/suppress most of the errors or fix them popping to PHP 8.0.

We are going to support PHP 7.4 on `2.16.x` branch until needed, for security patches only.

Ocramius:
TBH, dropping PHP version support, especially 7.4, isn't a major problem, since it has only 3 months of shelf life left.


Remember that `2.16.x` will still receive security issues until PHP 7.4 is dead.

PR #204, opened 2022-08-06
Apply PHP 8.0 syntax and property promotion

samsonasik:

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since we use php 8.0 as minimal requirement, we can apply php 8.0 syntax and property promotion.

### DIFF
--- a/src/Header/AbstractAddressList.php
+++ b/src/Header/AbstractAddressList.php
@@ -15,8 +15,8 @@ use function preg_match;
 use function preg_match_all;
 use function preg_replace;
 use function sprintf;
+use function str_contains;
 use function str_replace;
-use function strpos;
 use function strtolower;
 use function trim;
 
@@ -94,16 +94,13 @@ abstract class AbstractAddressList implements HeaderInterface
 
         $wasEncoded = false;
         $addresses  = array_map(
-            function ($value) use (&$wasEncoded) {
+            static function ($value) use (&$wasEncoded): ?Address {
                 $decodedValue = HeaderWrap::mimeDecodeValue($value);
                 $wasEncoded   = $wasEncoded || ($decodedValue !== $value);
-
-                $value = trim($decodedValue);
-
-                $comments = self::getComments($value);
-                $value    = self::stripComments($value);
-
-                $value = preg_replace(
+                $value        = trim($decodedValue);
+                $comments     = self::getComments($value);
+                $value        = self::stripComments($value);
+                $value        = preg_replace(
                     [
                         '#(?<!\\\)"(.*)(?<!\\\)"#', // quoted-text
                         '#\\\([\x01-\x09\x0b\x0c\x0e-\x7f])#', // quoted-pair
@@ -114,7 +111,6 @@ abstract class AbstractAddressList implements HeaderInterface
                     ],
                     $value
                 );
-
                 return empty($value) ? null : Address::fromString($value, $comments);
             },
             $values
@@ -183,7 +179,7 @@ abstract class AbstractAddressList implements HeaderInterface
             $name  = $address->getName();
 
             // quote $name if value requires so
-            if (! empty($name) && (false !== strpos($name, ',') || false !== strpos($name, ';'))) {
+            if (! empty($name) && (str_contains($name, ',') || str_contains($name, ';'))) {
                 // FIXME: what if name contains double quote?
                 $name = sprintf('"%s"', $name);
             }
--- a/src/Message.php
+++ b/src/Message.php
@@ -305,11 +305,9 @@ class Message
     /**
      * setSender
      *
-     * @param mixed $emailOrAddress
-     * @param mixed $name
      * @return Message
      */
-    public function setSender($emailOrAddress, $name = null)
+    public function setSender(mixed $emailOrAddress, mixed $name = null)
     {
         /** @var Sender $header */
         $header = $this->getHeaderByName('sender', Sender::class);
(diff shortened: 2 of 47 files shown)


samsonasik:
All green 🎉

src/Storage/AbstractStorage.php
```diff
@@ -232,12 +232,10 @@ public function offsetGet($id)
     /**
      * ArrayAccess::offsetSet()
      *
-     * @param mixed $id
-     * @param mixed $value
      * @throws Exception\RuntimeException
      */
     #[ReturnTypeWillChange]
-    public function offsetSet($id, $value)
+    public function offsetSet(mixed $id, mixed $value)
```
Ocramius, line 238: We cannot change types of public functions in non-`final` classes

src/Message.php
```diff
@@ -305,11 +305,9 @@ public function getReplyTo()
     /**
      * setSender
      *
-     * @param mixed $emailOrAddress
-     * @param mixed $name
      * @return Message
      */
-    public function setSender($emailOrAddress, $name = null)
+    public function setSender(mixed $emailOrAddress, mixed $name = null)
```
Ocramius, line 310: We cannot change this, if the class isn't `final`

src/Header/AbstractAddressList.php
```diff
@@ -183,7 +179,7 @@ public function getFieldValue($format = HeaderInterface::FORMAT_RAW)
             $name  = $address->getName();
 
             // quote $name if value requires so
-            if (! empty($name) && (false !== strpos($name, ',') || false !== strpos($name, ';'))) {
+            if (! empty($name) && (str_contains($name, ',') || str_contains($name, ';'))) {
```
Ocramius, line 182: Is the `empty()` check redundant here, or could it be `null`? 🤔

src/Storage/Folder.php
```diff
@@ -49,12 +29,16 @@ class Folder implements RecursiveIterator
...
      */
-    public function __construct($localName, $globalName = '', $selectable = true, array $folders = [])
-    {
-        $this->localName  = $localName;
+    public function __construct(
+        protected $localName,
+        $globalName = '',
+        protected $selectable = true,
+        /**
+         * subfolders of folder array(localName => \Laminas\Mail\Storage\Folder folder)
```
Ocramius: This should be in the method docblock

src/Storage/Folder.php
```diff
@@ -173,7 +157,7 @@ public function __unset($name)
      *
      * @return string global name of folder
      */
-    public function __toString()
+    public function __toString(): string
```
Ocramius, line 158: Adding a type here could break inheritance

src/Storage/Part.php
```diff
@@ -411,7 +411,7 @@ public function __isset($name)
      *
      * @return string content
      */
-    public function __toString()
+    public function __toString(): string
```
Ocramius, line 414: Adding a type declaration can break inheritance here

Ocramius (changes requested):

src/Storage/Folder.php
samsonasik: No; `__toString` with `Stringable` is special, extended class allowed to not have return type

src/Message.php
samsonasik: No, `mixed` type is special, and can be anything, if extended class method is typed, it will already error even on the beginning before it has `mixed` type

Ocramius: Oh, interesting, didn't know 🤔

src/Storage/AbstractStorage.php
Ocramius: In this case, isn't the ArrayAccess API stricter anyway? 🤔

src/Storage/Folder.php
samsonasik: ok, comment moved to method docblock https://github.com/laminas/laminas-mail/pull/204/commits/f258ba091963a1b702c7cb5353a65ea51a2966d4

src/Storage/Folder.php
```diff
@@ -4,57 +4,38 @@
...
+     * @param string $localName  local name (name of folder in parent folder)
      * @param string $globalName absolute name of folder
      * @param bool $selectable if true folder holds messages, if false it's
      *     just a parent for subfolders (Default: true)
-     * @param array<string, Folder> $folders init with given instances of Folder as subfolders
-     */
-    public function __construct($localName, $globalName = '', $selectable = true, array $folders = [])
-    {
-        $this->localName  = $localName;
+     * @param array<string, Folder> $folders subfolders of folder array(localName => \Laminas\Mail\Storage\Folder folder)
```
Ocramius: I think half this comment is unnecessary, given the type

src/Storage/Folder.php
Ocramius: Nice, didn't know 👍

src/Storage/Part.php
Ocramius: Similar to other comments in this patch, type is fine here 👍

Ocramius (approved):
LGTM, once green 👍

src/Storage/AbstractStorage.php
Ocramius: Meh, let's leave it for now 👍

src/Header/AbstractAddressList.php
Ocramius: This can be ignored for now 👍

Ocramius:
Thanks @samsonasik!

PR #227, opened 2023-01-10
Added auth plugin for Xoauth2

EmeryEx:
- Are you adding something the library currently does not support?
Yes
- Why should it be added? 
So you can connect to services like Microsoft Office 365 and utilize SMTP that doesn't support legacy protocols
- What will it enable? 
Xoauth2 auth protocol for SMTP
- How will the code be used? 
via the Laminas\Mail\Protocol\SmtpPluginManager as "xoauth2" or by "connection_class" option as "xoauth2" on SmtpOptions fed into SmtpTransport in a factory

I'm sorry i don't see a develop branch. I'm new to this.

### DIFF
--- a/src/Protocol/Smtp/Auth/Xoauth2.php
+++ b/src/Protocol/Smtp/Auth/Xoauth2.php
@@ -11,7 +11,7 @@ use function is_array;
 /**
  * Performs PLAIN authentication
  */
-class Xoauth2 extends Smtp
+final class Xoauth2 extends Smtp
 {
     /**
      * PLAIN username
--- a/src/Protocol/Smtp/Auth/Xoauth2.php
+++ b/src/Protocol/Smtp/Auth/Xoauth2.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Laminas\Mail\Protocol\Smtp\Auth;
+
+use Laminas\Mail\Protocol\Smtp;
+use Laminas\Mail\Protocol\Xoauth2\Xoauth2 as Xoauth2AuthEncoder;
+
+use function array_replace_recursive;
+use function base64_encode;
+use function is_array;
+
+/**
+ * Performs PLAIN authentication
+ */
+class Xoauth2 extends Smtp
+{
+    /**
+     * PLAIN username
+     *
+     * @var string
+     */
+    protected $username;
+
+    /**
+     * PLAIN password
+     *
+     * @var string
+     */
+    protected $accessToken;
+
+    /**
+     * @param  string $host   (Default: 127.0.0.1)
+     * @param  int    $port   (Default: null)
+     * @param  array  $config Auth-specific parameters
+     */
+    public function __construct($host = '127.0.0.1', $port = null, $config = null)
+    {
+        // Did we receive a configuration array?
+        $origConfig = $config;
+        if (is_array($host)) {
+            // Merge config array with principal array, if provided
+            if (is_array($config)) {
+                $config = array_replace_recursive($host, $config);
+            } else {
+                $config = $host;
+            }
+        }
+
+        if (is_array($config)) {
+            if (isset($config['username'])) {
+                $this->setUsername($config['username']);
+            }
+            if (isset($config['access_token'])) {
+                $this->setAccessToken($config['access_token']);
+            }
+        }
+
+        // Call parent with original arguments
+        parent::__construct($host, $port, $origConfig);
+    }
+
+    /**
+     * Perform XOAUTH2 authentication with supplied credentials
+     */
+    public function auth()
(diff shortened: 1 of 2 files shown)


EmeryEx:
This was a copy of AUTH PLAIN. Some of the checks are not passing for code that is already in the framework in same fashion in another file, but i've tried this on local and it works.

src/Protocol/Smtp/Auth/Xoauth2.php
```diff
@@ -0,0 +1,119 @@
...
+     * @var string
+     */
+    protected $accessToken;
+
+    /**
+     * @param  string $host   (Default: 127.0.0.1)
+     * @param  int    $port   (Default: null)
+     * @param  array  $config Auth-specific parameters
+     */
+    public function __construct($host = '127.0.0.1', $port = null, $config = null)
```
Slamdunk:
```suggestion
    public function __construct($host = '127.0.0.1', $port = null, ?array $config = null)
```
Signature should match parent's one

src/Protocol/Smtp/Auth/Xoauth2.php
```diff
@@ -0,0 +1,119 @@
...
+use Laminas\Mail\Protocol\Smtp;
+use Laminas\Mail\Protocol\Xoauth2\Xoauth2 as Xoauth2AuthEncoder;
+
+use function array_replace_recursive;
+use function is_array;
+
+/**
+ * Performs PLAIN authentication
+ */
+class Xoauth2 extends Smtp
```
Slamdunk:
```suggestion
final class Xoauth2 extends Smtp
```

Slamdunk (changes requested):
I see the Psalm errors are shared among all Auth classes, and you just copy-pasted most of the code.

I'm fine with this, but please update the psalm baseline to get the CI green

Slamdunk:
Thank you @EmeryEx

froschdesign:
@EmeryEx @Slamdunk 
There is a simple rule: what is not documented does not exist!
How can developers discover this and use it without documentation?

_It would be a pity if the work invested here is not used!_

Slamdunk:
My bad, sorry.

@EmeryEx would you be so kind to update https://github.com/laminas/laminas-mail/blob/2.22.x/docs/book/transport/smtp-authentication.md as well?

EmeryEx:
Yes I will add doc section in the next week or so when I get some time. Thank you guys!

liowalter:
Thanks for this @EmeryEx. Could you explain how to use it ? Where do I get the access token ?

EmeryEx:
> Thanks for this @EmeryEx. Could you explain how to use it ? Where do I get the access token ?

It's no easy task and unfortunately, it's something you need to implement yourself. But once you have the access token with the right permissions, you just have to supply it to SmtpTransport:

```
$transport = new SmtpTransport();

$connection_conf = [
    'username' => $username, // the email address of user that approved token
    'access_token' => $token, // the access token you've acquired via an authorization token or refresh token reuest
    'ssl' => 'tls',
];

$options   = new SmtpOptions([
    'name'              => 'example.com',
    'host'              => 'smtp.office365.com',
    'port'              => 587,
    'connection_class'  => 'Xoauth2',
    'connection_config' => $connection_conf,
]);
        
$transport->setOptions($options);
```

Then your transport is ready to send mail.

Here are the docs for Office365 on obtaining Access tokens and such:
[Docs](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code)

One more thing after you've acquired an access token, you still need to enable SMTP auth for the email account as an admin (one time)
[Enable SMTP authentication for Office 365 account](https://learn.microsoft.com/en-us/exchange/clients-and-mobile-in-exchange-online/authenticated-client-smtp-submission#use-the-microsoft-365-admin-center-to-enable-or-disable-smtp-auth-on-specific-mailboxes)

froschdesign:
@EmeryEx 
Please create a pull request for it to update the documentation.

EmeryEx:
> @EmeryEx Please create a pull request for it to update the documentation.

Where exactly should i put it? Not sure the docs flow you've got going on

froschdesign:
@EmeryEx 
In the documentation you will find a [chapter "SMTP authentication"](https://docs.laminas.dev/laminas-mail/transport/smtp-authentication/), add it there.
The related file can be found in the `docs` folder of this repository:

https://github.com/laminas/laminas-mail/blob/2.22.x/docs/book/transport/smtp-authentication.md

YalagureshJ:
Hi @EmeryEx, Is this working? Please share the documentation on how to use it?

PR #235, opened 2023-05-03
Update Sendmail.php

### LINKED ISSUE #230: Use "text/html" content type instead of "multipart/alternative"
### Feature Request

Use "text/html" content type instead of "multipart/alternative" [in some cases].

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Currently Laminos Mail sends all messages in "multipart/alternative" format that makes them not readable by some e-mail clients.
A quote from mail source:

```
Content-Type: multipart/alternative;
...
This is a message in Mime Format.  If you see this, your mail reader does not support this format.
```

But some products do not need to send mails with a complicated structure, they need just to send text or HTML messages.
Here is the issue description on osTicket forum (they use LM lib): https://forum.osticket.com/d/102563-ya-this-is-a-message-in-mime-format/18

It could be nice if LM uses "text/html" content type as all other SMTP clients do. Maybe not mandatory, but using some option.

cloudseeder:
Remove the \r\n translation for mail headers per the PHP mail() page. 

additional_headers (optional)
----------
String or array to be inserted at the end of the email header.

This is typically used to add extra headers (From, Cc, and Bcc). Multiple extra headers should be separated with a CRLF (\r\n). If outside data are used to compose this header, the data should be sanitized so that no unwanted headers could be injected.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This is fully documented in osTicket BR https://github.com/laminas/laminas-mail/issues/230.


Slamdunk:
Hi, can you add a test for the bugfix please?

cloudseeder:
@Slamdunk

Sure. Can you point me to an example? Looks like there are a lot of good examples in MessageTest.php. Should I just copy/mod one of those tests? With all those tests I'm really surprised this wasn't caught earlier.

So what I think I need to do is:

1. Edit MessageTest.php
2. Copy and modify one of the existing tests to meet the needs of this PR
3. Ensure it works
4. Add the test to this PR (is that correct?) 

I added a new file into my fork ([cloudseeder/laminas-mail/tree/2.23.x/test)/Storage/HeaderTest.php) and some ci event kicked off from the main branch. Just want to make sure I do this correctly. I'm more of a ops guy than a programmer.

Slamdunk:
That's correct: go for it :muscle:

PR #238, opened 2023-05-19
Adds static keyword to getPluginManagerMethod

schnumsel:
PHPUnit Tests fail because of [BC break in ServiceManager 3.21](https://github.com/laminas/laminas-servicemanager/pull/184/files#diff-5470c7fee9ea3d84104b00ba271bfd93257384018f64b1a57e8105fc8b06ea2d)

This PR tries to fix the issue


schnumsel:
Well, this is impossible to solve (for me), since some of the test run with ServiceManager version 3.20 (locked) and others (latest) with version 3.21...
Giving up here

Ocramius:
@schnumsel now fixed in mainline

PR #241, opened 2023-05-22
Type of returning for functions of the prepareSubject and the prepare…

### LINKED ISSUE #240: Laminas\Mail\Transport\Exception\RuntimeException: Unable to send mail: mail(): Passing null to parameter #2 ($subject) of type string is deprecated in /app/vendor/laminas/laminas-mail/src/Transport/Sendmail.php:310
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | x.y.z

artemii-karkusha:
Type of returning for functions of the prepareSubject and the prepareParameters have been fixed

#240


Ocramius (changes requested):
We'd need a test to verify this change: I suppose you expect `getSubject()` to always return a `string`?

Xerkus:
dockblock and usage checks out, it is expected to always return string.

Return values are passed to `mail()` with signature
```
 mail(
    string $to,
    string $subject,
    string $message,
    array|string $additional_headers = [],
    string $additional_params = ""
): bool
```

Ocramius:
Yes, I mean a test case (additions to our PHPUnit test suite :D )

The test should fail before this change, and pass after this change.

artemii-karkusha:
> We'd need a test to verify this change: I suppose you expect `getSubject()` to always return a `string`?

I will add for this case

artemii-karkusha:
@Ocramius 
Result of test the testAllowMessageWithEmptySubjectButHasBccHeader before changes for SendMail was made. 
Screenshot 2023-05-23 at 11 39 26

test/Transport/SendmailTest.php
```diff
@@ -342,8 +342,30 @@ public function testDoesNotInjectFromParameterFromSenderWhenFromOptionPresentInP
...
+    {
+        $message = Message::fromString('')->setEncoding('utf-8');
+        $message
+            ->addTo('matthew@example.org')
+            ->addFrom('ralph@example.org')
+            ->setBody("Sorry, I'm going to be late today!");
+        $this->transport->setCallable([$this->transport, 'mailHandler']);
+        try {
+            $this->transport->send($message);
+        } catch (\RuntimeException $runtimeException) {
```
Ocramius: Is a `RuntimeException` expected to be thrown at all times?

If so, PHPUnit can help you here:

```php
$this->expectException(\RuntimeException::class);
$this->expectExceptionMessage('Something different from a warning');
```

The try-catch can then be removed :)

Ocramius (changes requested):
@artemii-karkusha looking much better now, thanks!

What's missing now is some cleanup around the test case.

test/Transport/SendmailTest.php
artemii-karkusha: @Ocramius No. It is not expected to throw all time. It was added to be sure logic will not be broken for mailHandler in future.

artemii-karkusha:
@Ocramius Please let me know if something is needed to changing. It is ready for merging

test/Transport/SendmailTest.php
Ocramius: Ok, but then this test has no assertions, and is a risky test :thinking: 

Can we somehow verify the outcome with a valid assertion?

/cc @Slamdunk for ideas

artemii-karkusha: It is not real to cover outcome. Function doesn't return nothing. What we can cover is exception which is thrown when someone will change behaviour for subject and it will influence on function the mail(..., #subject). 
prepareSubject which can be better place for test case is protected.
I don't have any ideas how can we cover it in another way. If you have let it will share with me, please.

Ocramius: Yep, that's why I poked @Slamdunk to get some more brains :)

Slamdunk: IMHO we don't need tests here, it's a type mismatch so getting rid of PSalm suppressions should be enough:

```diff
diff --git a/psalm-baseline.xml b/psalm-baseline.xml
index 88d83696..0140e43e 100644
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1737,18 +1737,12 @@
     <InvalidArgument>
       <code><![CDATA[[$this, 'handleMailErrors']]]></code>
     </InvalidArgument>
-    <InvalidNullableReturnType>
-      <code>string</code>
-    </InvalidNullableReturnType>
     <MissingReturnType>
       <code>mailHandler</code>
     </MissingReturnType>
     <MixedAssignment>
       <code>$param</code>
     </MixedAssignment>
-    <MixedInferredReturnType>
-      <code>string</code>
-    </MixedInferredReturnType>
     <MixedOperand>
       <code>$param</code>
     </MixedOperand>
diff --git a/src/Transport/Sendmail.php b/src/Transport/Sendmail.php
index aae76f49..58868dda 100644
--- a/src/Transport/Sendmail.php
+++ b/src/Transport/Sendmail.php
@@ -9,6 +9,7 @@ use Laminas\Mail\Transport\Exception\InvalidArgumentException;
 use Laminas\Mail\Transport\Exception\RuntimeException;
 use Traversable;
 
+use function assert;
 use function count;
 use function escapeshellarg;
 use function gettype;
@@ -200,10 +201,12 @@ class Sendmail implements TransportInterface
     {
         $headers = $message->getHeaders();
         if (! $headers->has('subject')) {
-            return;
+            return '';
         }
-        $header = $headers->get('subject');
-        return $header->getFieldValue(HeaderInterface::FORMAT_ENCODED);
+        $header     = $headers->get('subject');
+        $fieldValue = $header->getFieldValue(HeaderInterface::FORMAT_ENCODED);
+        assert(is_string($fieldValue));
+        return $fieldValue;
     }
 
     /**
@@ -259,7 +262,7 @@ class Sendmail implements TransportInterface
     protected function prepareParameters(Mail\Message $message)
     {
         if ($this->isWindowsOs()) {
-            return;
+            return '';
         }
 
         $parameters = (string) $this->parameters;
```

Ocramius: I'm happy with this outcome too :+1:

artemii-karkusha: @Ocramius 
It has to be closed. I did mistake when changed author for commits. I have created new PL with all needed changes. 
https://github.com/laminas/laminas-mail/pull/242

Slamdunk:
See #242

PR #242, opened 2023-05-24
Fix return-type for `Sendmail::prepareSubject` and the `Sendmail::prepareParameters`

### LINKED ISSUE #240: Laminas\Mail\Transport\Exception\RuntimeException: Unable to send mail: mail(): Passing null to parameter #2 ($subject) of type string is deprecated in /app/vendor/laminas/laminas-mail/src/Transport/Sendmail.php:310

artemii-karkusha:
Type of returning for functions of the prepareSubject and the prepareParameters have been fixed

https://github.com/laminas/laminas-mail/issues/240

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1663,6 +1663,9 @@
     </MissingReturnType>
   </file>
   <file src="src/Transport/Envelope.php">
+    <InvalidNullableReturnType>
+      <code>string</code>
+    </InvalidNullableReturnType>
     <MissingReturnType>
       <code>setFrom</code>
       <code>setTo</code>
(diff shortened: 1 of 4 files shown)


artemii-karkusha:
@Ocramius 
Suggested changes have been applied.. Test case has been reverted. 
I have change the  Laminas\Mail\Transport\Envelope::getFrom, because of issue with type of property the $from. 
Thank you.

psalm-baseline.xml
```diff
@@ -1663,9 +1663,6 @@
     </MissingReturnType>
   </file>
   <file src="src/Transport/Envelope.php">
-    <InvalidNullableReturnType>
-      <code>string</code>
-    </InvalidNullableReturnType>
```
Slamdunk: I see you've removed the `InvalidNullableReturnType` from `src/Transport/Envelope.php`, but I rather not touch the Envelope class properties as it could be seen as a BC Break.

In https://github.com/laminas/laminas-mail/pull/241#discussion_r1203676239 the only `InvalidNullableReturnType` removal I suggest is the `src/Transport/Sendmail.php`, which in this PR is still in place

Slamdunk (changes requested):

Slamdunk (approved):

Slamdunk:
Thank you

PR #251, opened 2023-11-01
Remove dependency on laminas-crypt

gsteel:
Instead of requiring laminas-crypt, call `hash_hmac` directly.

A few psalm issues have been base-lined because we can't add native types without breaking BC. Also, some defensive conditions contradict doc-block types.

Subtle potential BC break here is that an exception will now be thrown if the password is empty|null, whereas previously, it would likely have been a type error.

Also see: https://github.com/laminas/laminas-crypt/pull/32

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,16 +20,14 @@
     },
     "require-dev": {
         "laminas/laminas-coding-standard": "~2.5.0",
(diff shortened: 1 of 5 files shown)


Slamdunk:
Excellent work :muscle: 

Thank you @gsteel

PR #264, opened 2024-05-31
Update ContentType.php

### LINKED ISSUE #119: Laminas mail expects a value in the format "type/subtype"; received "image/*"
Issue detected when using GLPI Mailgate https://github.com/glpi-project/glpi/issues/8375 to get messages with attached files "image/*"

To fix this issue, the RegEx at the file "laminas-mail/src/Header/ContentType.php", line 114 should be changed to

/^[a-z-]+/[/*a-z0-9.+-]+$/i

tomolimo:
To be able to cope with mimetype like type/*

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

fixes #119 

eml file to test this bug is here: https://gist.github.com/rikvdh/489e13b22f9012db2dcd9f5046b53e43

### DIFF
--- a/src/Header/ContentType.php
+++ b/src/Header/ContentType.php
@@ -131,7 +131,7 @@ class ContentType implements UnstructuredInterface
      */
     public function setType($type)
     {
-        if (! preg_match('/^[a-z-]+\/[a-z0-9.+-]+$/i', $type)) {
+        if (! preg_match('/^[a-z-]+\/[\*a-z0-9.+-]+$/i', $type)) {
             throw new Exception\InvalidArgumentException(sprintf(
                 '%s expects a value in the format "type/subtype"; received "%s"',
                 __METHOD__,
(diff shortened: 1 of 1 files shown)


src/Header/ContentType.php
```diff
@@ -131,7 +131,7 @@ public function toString()
      */
     public function setType($type)
     {
-        if (! preg_match('/^[a-z-]+\/[a-z0-9.+-]+$/i', $type)) {
+        if (! preg_match('/^[a-z-]+\/[\*a-z0-9.+-]+$/i', $type)) {
```
Xerkus: Escape sequence is not needed for `*`. It does not have special meaning in character group.

tomolimo: Hello @Xerkus 
Yes, you are right, I'm going to delete the escape char
Thank you
Regards,
Tomolimo

Xerkus:
This change will need an accompanying unit test. Both in reading content type from string and returning expected string representation with `*`

tomolimo:
Unfortunately, I'will not have time to do it...
Sorry for that

